PR #177, opened 2026-03-27
fix(dialyzer): configure dialyzer + resolve 133 errors

jat10:


jfayad (changes requested):
@jat10 are you sure we want to enforce the _ = XXX format ?

AFAIK it's not a best practice and it's adding a lot of unnecessary verbosity to the code

>Dialyxir does not enable that warning by default in modern versions; if you are seeing it, your project likely has :unmatched_returns enabled in mix.exs, or you are passing --unmatched_returns on the CLI. The Dialyxir docs say warning flags are optional and passed through to Dialyzer, and they show :unmatched_returns / "-Wunmatched_returns" as an example opt-in flag.  ￼
>
>To disable it, remove that flag from your config.

I'l also surprised to see no changes to mix.exs with a reference to dialixyr (if this is what you used)

PR #219, opened 2026-04-05
feat(llm): added anthropic llm

Arkhamides:
## Summary
Add Anthropic provider support to allow using Claude models alongside OpenAI, plus fix form submission in the chat interface.

## Changes
- **Anthropic LLM Support**: Added provider-aware configuration to support both OpenAI and Anthropic APIs
- Routes to correct endpoints based on provider
- Instantiates appropriate LangChain models (ChatAnthropic or ChatOpenAI)
- Updated system config UI to handle both providers

- **Form Submission Fix**: Replaced manual submit event dispatching with `requestSubmit()` for proper form handling. (Previously pressing enter would refresh the browser and make the chat disapear)

### DIFF
--- a/lib/zaq/agent/llm.ex
+++ b/lib/zaq/agent/llm.ex
@@ -88,14 +88,10 @@ defmodule Zaq.Agent.LLM do
       model: cfg.model,
       temperature: cfg.temperature,
       top_p: cfg.top_p,
-      endpoint: endpoint_for(cfg),
+      endpoint: cfg.endpoint <> cfg.path,
       api_key: cfg.api_key
     }
 
     Map.merge(base, Map.new(overrides))
   end
-
-  defp endpoint_for(%{provider: "anthropic", endpoint: ep}) when ep not in [nil, ""], do: ep
-  defp endpoint_for(%{provider: "anthropic"}), do: "https://api.anthropic.com/v1/messages"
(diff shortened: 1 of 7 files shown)


lib/zaq/agent/llm.ex
```diff
@@ -84,13 +84,18 @@ defmodule Zaq.Agent.LLM do
...
+      endpoint: endpoint_for(cfg),
       api_key: cfg.api_key
     }
 
     Map.merge(base, Map.new(overrides))
   end
+
+  defp endpoint_for(%{provider: "anthropic", endpoint: ep}) when ep not in [nil, ""], do: ep
+  defp endpoint_for(%{provider: "anthropic"}), do: "https://api.anthropic.com/v1/messages"
+  defp endpoint_for(%{endpoint: ep}), do: ep <> "/chat/completions"
```
jfayad: The base_url for anthropic is already supported by the models' DB we are using.
The path is what needs to be dynamically provided based on the provider.

tiredbasil96:
Capture d’écran 2026-04-06 à 12 07 23

Hello, thank you for the PR!

I checked the impact of your PR over the form and noticed the drop-down functionality for the Model input is gone. This is a regression. It must be kept to facilitate the choice for the users.

jfayad (changes requested):

jat10 (approved):

jfayad (approved):

PR #266, opened 2026-04-17
fix(ingestion): deduplicate uploaded filenames using OS-style (n) suffix

### LINKED ISSUE #262: fix(ingestion): file upload management
When we upload a file with the same name, the process status is added to it, which is wrong. To fix this, we need to have a naming convention like OS handles the same file name by adding (1)

jat10:
Fix #262


jat10:
This has been tested:
- Upload file `test.png`
- Fail the ingestion on purpose to show the error tag
- Upload same file it created `test(1).png` and no error tags 

Screenshot 2026-04-17 at 7 14 10 PM

PR #296, opened 2026-04-22
fix(retrieval): disable json mode when conversation history is present

jat10:


lib/zaq/agent/retrieval.ex
```diff
@@ -89,8 +91,11 @@ defmodule Zaq.Agent.Retrieval do
     end
   end
 
-  defp maybe_add_json_mode(config) do
-    if LLM.supports_json_mode?() do
+  # Some providers (e.g. Novita) return null content when json_response is active
+  # and history contains non-JSON assistant messages. Disabling JSON mode when
+  # history is present is safe — extract_json/1 handles JSON embedded in free-form text.
+  defp maybe_add_json_mode(config, history) do
+    if LLM.supports_json_mode?() and history == [] do
```
jfayad, line 98: looks like a workaround not a fix

PR #316, opened 2026-04-30
feat(agent): replace on_status closure with Status.broadcast and gate PromptGuard in Api

### LINKED ISSUE #279: [BO Chat] Wire New Agent's call into BO updates
- [x] Fix the wiring of the calls to the on_status: Event.opts through NodeRouter

on_status: WebBridge.on_status_callback(session_id, request_id),

This wiring is wrong.

Moreover we shouldn't invoke the webBridge directly, every outgoing call should be sent to the Channels.Router

- [x] Once this is done, wire the calls to update the live answering bot status when the Agent.API routed the call through a custom agent
- [x] Display tool calls during the answer generation
- [x] Pass the Zaq.Event struct down the agents and down to the Actions/tools they execute

jat10:
Fix #279


lib/zaq/agent/status.ex
```diff
@@ -0,0 +1,54 @@
...
+
+  defp broadcast_to_session(session_id, request_id, stage, message, node_router) do
+    Event.new(
+      %{
+        module: Phoenix.PubSub,
+        function: :broadcast,
+        args: [Zaq.PubSub, "chat:#{session_id}", {:status_update, request_id, stage, message}]
+      },
+      :bo
+    )
```
jfayad, line 73: if you don't care about the return, why hang execution here ?

Send the `type: :async` opt as a third param to the new/3 function

(I double checked and noticed the dispatch now doesn't support async, I'm working on that in a separate fix PR)

jfayad (changes requested):

jfayad:
@jat10 async EventHop types is now supported in the noderouter dispatch

jfayad (approved):

PR #320, opened 2026-05-01
fix(release): fix llm_db dep

jat10:
Problem:

The Docker file in the build is calling mix release, it was failing with the following error:

```
** (Mix) :llm_db is listed both as a regular application and as an included application
```

### DIFF
--- a/.github/workflows/elixir-ci.yml
+++ b/.github/workflows/elixir-ci.yml
@@ -77,6 +77,8 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
+    permissions:
+      statuses: write
 
     services:
       postgres:
--- a/.github/workflows/elixir-ci.yml
+++ b/.github/workflows/elixir-ci.yml
@@ -77,8 +77,6 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
-    permissions:
-      statuses: write
 
     services:
       postgres:
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # ZAQ
 
 [![Coverage Status](https://img.shields.io/coveralls/github/www-zaq-ai/zaq?branch=main)](https://coveralls.io/github/www-zaq-ai/zaq?branch=main)
+[![Discord](https://img.shields.io/discord/1488938719959322624?color=7289da&logo=discord&logoColor=white&label=discord)](https://discord.gg/rDUeWP5GbD)
 [![Docs](https://img.shields.io/badge/docs-github%20pages-blue)](https://www-zaq-ai.github.io/zaq/)
 [![Run](https://img.shields.io/badge/quick%20run-local%20setup-orange)](https://github.com/www-zaq-ai/zaq/wiki/Local-Installation)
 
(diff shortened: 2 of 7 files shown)


jat10:
Also fixed the flaky test `test/zaq/agent/server_manager_test.exs:1204 `

lib/zaq/application.ex
```diff
@@ -58,6 +59,7 @@ defmodule Zaq.Application do
     case Supervisor.start_link(children, opts) do
       {:ok, _pid} = ok ->
         enqueue_release_badge_check_on_startup()
+        _ = ValidModalities.list()
```
jfayad, line 64: what's this ?

.github/workflows/elixir-ci.yml
```diff
@@ -77,6 +77,8 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
+    permissions:
+      statuses: write
```
jfayad: as we saw, coveralls is back and this change isn't needed so please remove it

config/dev.exs
```diff
@@ -115,12 +115,12 @@ config :zaq, ZaqWeb.Endpoint,
     web_console_logger: true,
     patterns: [
       # Static assets, except user uploads
-      ~r"priv/static/(?!uploads/).*\.(js|css|png|jpeg|jpg|gif|svg)$",
+      ~r"priv/static/(?!uploads/).*\.(js|css|png|jpeg|jpg|gif|svg)$"E,
```
jfayad, line 118: what's this for ?

test/zaq/agent/server_manager_test.exs
jfayad, line 1: you can remove that workaround, these flakiness have been solved on main now

jfayad (changes requested):

config/dev.exs
jat10: @jfayad these to make the mix release work on dev. currently if you run mix release it fails

.github/workflows/elixir-ci.yml
jat10: noted

test/zaq/agent/server_manager_test.exs
jat10: noted

README.md
```diff
@@ -1,6 +1,7 @@
 # ZAQ
 
 [![Coverage Status](https://img.shields.io/coveralls/github/www-zaq-ai/zaq?branch=main)](https://coveralls.io/github/www-zaq-ai/zaq?branch=main)
+[![Discord](https://img.shields.io/discord/1488938719959322624?color=7289da&logo=discord&logoColor=white&label=discord)](https://discord.gg/rDUeWP5GbD)
```
jfayad: the rendering of that badge seems incomplete, can you fix both the color and the text ?

see preview here: https://github.com/www-zaq-ai/zaq/blob/f6915e8403f154831b856c92676e96f85d1fa155/README.md

jfayad (changes requested):
can you increase code coverage instead of dropping it ? :-)

README.md
jat10: The widget's display purpose is to display the number of members, but since we are only 10 now, we will update that later

Check it here: https://github.com/www-zaq-ai/zaq/blob/fix/release/README.md

jfayad (approved):

PR #347, opened 2026-05-05
fix(ingest): update document sources and sidecar metadata when renaming a folder

### LINKED ISSUE #331: [Ingestion]: Renaming a folder
When the user renames the folder, the sidecar and file permissions are lost. 

The folder i have called zaq:

Image

Rename it to product:

Image

jat10:
fix #331


lib/zaq/ingestion/document.ex
```diff
@@ -98,6 +98,56 @@ defmodule Zaq.Ingestion.Document do
     Repo.delete(document)
   end
 
+  @doc """
+  Bulk-renames document sources that begin with `old_prefix/` by replacing the
+  prefix with `new_prefix`. Used when a folder is renamed on the filesystem.
+  Returns `{count, nil}`.
+  """
+  def rename_source_prefix(old_prefix, new_prefix) do
```
jfayad: shouldn't these be an atomic transaction for all ?

jfayad (changes requested):

jat10:
@jfayad ready for review

test/zaq/ingestion/rename_service_test.exs
```diff
@@ -0,0 +1,114 @@
...
+
+      perm = Repo.get_by(Permission, document_id: updated_doc.id, person_id: person.id)
+      assert perm != nil, "Permission should survive the folder rename"
+      assert perm.access_rights == ["read"]
+    end
+
+    test "leaves filesystem and DB unchanged when rename fails", %{doc: doc} do
+      # Renaming to a path whose parent does not exist causes File.rename to
+      # return {:error, :enoent}, exercising the failure branch.
+      assert {:error, _} = RenameService.rename_entry("default", "zaq", "nonexistent/product")
```
jfayad, line 102: looks like this fails early on.

You need a test that guarantees rollback was done correctly even if the last operation of your Ecto.multi failed

lib/zaq/ingestion/rename_service.ex
```diff
@@ -14,7 +14,9 @@ defmodule Zaq.Ingestion.RenameService do
   end
 
   defp rename_by_type(:directory, volume_name, old_path, new_path, _volumes) do
-    FileExplorer.rename(volume_name, old_path, new_path)
+    with :ok <- FileExplorer.rename(volume_name, old_path, new_path) do
```
jfayad: Running this outside the transaction means the file name can be changed but the DB records won't be aligned with it anymore.

jfayad (changes requested):
This is a major operation and multiple features depend on a correct state of files

1. Test coverage is not enough
2. Inconsistent state seem possible with the current implementation

lib/zaq/ingestion/document.ex
```diff
@@ -98,6 +98,50 @@ defmodule Zaq.Ingestion.Document do
...
+    )
+  end
+
+  @doc """
+  Returns a query that bulk-renames one JSONB metadata key (e.g. `"sidecar_source"`)
+  whose value begins with `old_prefix/`. Intended for use inside an `Ecto.Multi`.
+  """
+  def rename_metadata_key_query(key, old_prefix, new_prefix) do
+    from(d in __MODULE__,
+      where: like(fragment("?->>?::text", d.metadata, ^key), ^"#{old_prefix}/%"),
```
jfayad: shouldn't this be ?
```elixir
fragment("?->>(?::text)", d.metadata, ^key)
```

lib/zaq/ingestion/rename_service.ex
```diff
@@ -25,23 +33,83 @@ defmodule Zaq.Ingestion.RenameService do
...
+  end
+
+  defp rename_with_saga(rename_ops, db_multi) do
+    saga =
+      Enum.reduce(rename_ops, Sage.new(), fn op, sage ->
+        Sage.run(
+          sage,
+          {:fs, op.volume, op.old_relative},
+          fn _, _ -> fs_rename_step(op.volume, op.old_relative, op.new_relative) end,
+          fn _, _, _ -> fs_compensate_step(op.volume, op.old_relative, op.new_relative) end
```
jfayad, line 48: duplicate fuctions, why not simply directly calling FileExplorer.rename (with inverted params) ?

jfayad (changes requested):

lib/zaq/ingestion/rename_service.ex
jat10: The linter's max-depth-2 rule is the reason those helpers need to exist

jfayad (approved):

PR #384, opened 2026-05-08
fix(docker-hub): add single all-in-one Docker image with ParadeDB and ZAQ bundled

### LINKED ISSUE #359: [local-installion]: one docker image
We need to create a single Docker that contains the zaq image and the DB

This will simplify the installation no docker-compose 

We need to leave a message to help the user add volumes in Docker Desktop.

jat10:
fix #359


jat10:
This will stay on hold until next release on 0.8.2 so it build the image on docker-hub. Also it changes the README

PR #430, opened 2026-05-27
feat(workflow): bo screens

jat10:


config/runtime.exs
```diff
@@ -20,9 +20,23 @@ if System.get_env("PHX_SERVER") do
...
+  finch: [
+    name: ReqLLM.Finch,
+    pools: %{
+      :default => [
+        protocols: [:http1],
+        size: 1,
+        count: String.to_integer(System.get_env("REQ_LLM_POOL_COUNT", "20"))
+      ]
+    }
+  ]
```
jfayad, line 35: what is this for ?

why not using http2 ?

lib/zaq/agent/tools/accounts/history.ex
```diff
@@ -0,0 +1,91 @@
...
+    ]
+
+  use Zaq.Engine.Workflows.Action
+
+  require Logger
+
+  alias Zaq.Engine.Conversations
+
+  @impl Jido.Action
+  def run(%{person_id: person_id} = params, _ctx) when not is_nil(person_id) do
```
jfayad: secure with permission access ?

lib/zaq/agent/tools/accounts/history.ex
```diff
@@ -0,0 +1,91 @@
...
+    conv_limit = params[:conversation_limit] || 10
+    msg_limit = params[:messages_per_conversation] || 50
+
+    history =
+      [person_id: person_id, limit: conv_limit]
+      |> Conversations.list_conversations()
+      |> Enum.flat_map(fn conv ->
+        conv
+        |> Conversations.list_messages()
+        |> Enum.take(msg_limit)
```
jfayad: can't you provide the limit to the list_messages/1 function so it gets limited on the DB call

lib/zaq/agent/tools/accounts/history.ex
```diff
@@ -0,0 +1,91 @@
...
+    )
+
+    {:ok, %{history: history}}
+  rescue
+    error ->
+      Logger.warning(
+        "[History] failed person_id=#{inspect(person_id)} error=#{Exception.message(error)}"
+      )
+
+      {:ok, %{history: []}}
```
jfayad: a failure should not return an empty history but the actual error so the agent can adjust and the workflow can fail

lib/zaq/agent/tools/email/draft_reply.ex
```diff
@@ -35,45 +35,82 @@ defmodule Zaq.Agent.Tools.Email.DraftReply do
...
+          %Incoming{
+            content: build_prompt(from_name || from_address, subject, body),
+            channel_id: from_address,
+            author_id: from_address,
+            author_name: from_name,
+            provider: :"email:imap",
+            metadata: %{"subject" => subject}
+          }
+
+        outgoing = executor.run(incoming, agent_id: agent_id, scope: scope)
```
jfayad: This should be dispatched through the NodeRouter so all verification made before run applies

Also why isn't this just composed of other re-usable actions ?

lib/zaq/agent/tools/email/draft_reply.ex
```diff
@@ -35,45 +35,82 @@ defmodule Zaq.Agent.Tools.Email.DraftReply do
...
+      per_email_logs ++
+        [
+          %{
+            level: "info",
+            message: "Drafted #{length(drafts)} reply(s) via agent '#{agent_name}'",
+            metadata: %{count: length(drafts), agent: agent_name}
+          }
+        ]
 
     {:ok, %{drafts: drafts}, logs: logs}
```
jfayad: what is the logs field for ?

lib/zaq/agent/tools/email/fetch_emails.ex
jfayad: according to this comment, why isn't this action removed ?

lib/zaq/agent/tools/email/notify_empty_mailbox.ex
jfayad, line 1: according to the comment in this action, this should be removed

lib/zaq/agent/tools/google_drive/create_document.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Zaq.Agent.Tools.CreateDocument do
+defmodule Zaq.Agent.Tools.GoogleDrive.CreateDocument do
```
jfayad: who asked you to rename this action ?

it is generic for any DataSource not for Google Drive specifically...

If you want to organize in folders, the folder name should be `data_source` -> `DataSource`

lib/zaq/agent/tools/data_source/download_document.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/data_source/get_document.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/data_source/list_documents.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/data_source/search_documents.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/data_source/update_document.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/sheets/add_sheet_tab.ex
jfayad, line 1: This action has been built to be compatible with any sheets provider not specifically GoogleSheets

lib/zaq/agent/tools/sheets/append_sheet_values.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/sheets/clear_sheet_values.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/sheets/create_sheet.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/sheets/delete_sheet_tab.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/google_sheets/extract_rows.ex
```diff
@@ -0,0 +1,85 @@
+defmodule Zaq.Agent.Tools.GoogleSheets.ExtractRows do
+  @moduledoc """
+  Workflow action: converts raw Google Sheet content into normalized row maps.
```
jfayad: 1st this isn't specific to google Sheets
2nd you are not normalizing a sheet doc, you are normalizing a row content type of a Record and you are assuming that the first row is for the header
3rd the record is not any type, it's a Record struct and it should be enforced

lib/zaq/agent/tools/sheets/get_sheet.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/sheets/get_sheet.ex
```diff
@@ -24,10 +24,16 @@ defmodule Zaq.Agent.Tools.GetSheet do
       spreadsheet_id: [type: :string, required: true, doc: "Spreadsheet identifier"],
       range: [type: :string, required: false, doc: "Optional A1 range"],
       config_id: [type: :string, required: false, doc: "Optional scoped datasource config id"]
+    ],
+    output_schema: [
+      record: [type: :any, required: true, doc: "Normalized spreadsheet record."]
```
jfayad: record is a known struct

lib/zaq/agent/tools/sheets/inspect_sheet.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/sheets/update_sheet_values.ex
```diff
@@ -36,10 +36,17 @@ defmodule Zaq.Agent.Tools.UpdateSheetValues do
         doc: "Provider value input option (for example USER_ENTERED)"
       ],
       config_id: [type: :string, required: false, doc: "Optional scoped datasource config id"]
+    ],
+    output_schema: [
+      status: [type: :string, required: true, doc: "Result status, e.g. \"updated\"."],
+      record: [type: :any, required: false, doc: "Datasource record returned by the provider."]
```
jfayad: record is a typed struct

lib/zaq/agent/tools/sheets/update_sheet_values.ex
jfayad, line 1: revert name change

lib/zaq/agent/tools/people/ensure_person.ex
```diff
@@ -1,81 +1,110 @@
...
+      email: [
+        type: :string,
+        required: false,
+        doc: "Email address; also used as channel_id for 'email' platform."
+      ],
+      phone: [type: :string, required: false, doc: "Phone number for matching."]
     ],
     output_schema: [
-      drafts: [type: {:list, :map}, required: true]
+      person_id: [type: :integer, required: false, doc: "ID of the found or created Person."],
```
jfayad: why not returning the actual Person struct ?

lib/zaq/agent/tools/people/notify_person.ex
```diff
@@ -0,0 +1,181 @@
...
+        %{}
+      )
+      # => {:ok, %{notified: true, channel_identifier: "jad@example.com", status: :dispatched}}
+  """
+
+  use Jido.Action,
+    name: "notify_person",
+    description: "Notify a person via a configured channel medium (email, mattermost, slack…).",
+    schema: [
+      person_id: [type: :any, required: true, doc: "Person ID (integer or string)."],
```
jfayad: why not taking the full person to already have all the required data

lib/zaq/agent/tools/people/notify_person.ex
```diff
@@ -0,0 +1,181 @@
...
+      medium: [
+        type: :string,
+        required: true,
+        doc: "Channel platform: email, mattermost, slack, etc."
```
jfayad: why is this required here ? shouldn't that be the responsibility of the Notification center (to pick the preferred channel and circle over other ones if the first one isn't available etc...)

lib/zaq/agent/tools/people/notify_person.ex
```diff
@@ -0,0 +1,181 @@
...
+  ### Optional sheet-tracking fields
+
+  When a subsequent workflow step needs to update a spreadsheet cell (e.g. to
+  increment an email-state counter), pass these and the computed `range` /
+  `values` will be included in the output for direct use by
+  `UpdateSheetValues`:
+
+  - `row_index`          — 1-based sheet row number.
+  - `email_state`        — current counter value; output will carry `email_state + 1`.
+  - `email_state_column` — column letter (default `"I"`).
```
jfayad: this is too specific and attached to a specific workflow, this does NOT belong here

lib/zaq/agent/tools/pipeline_runner.ex
jfayad, line 1: what is this for ?

lib/zaq/agent/tools/workflow/sleep.ex
jfayad, line 1: why did you need to redeclare this action which is already provided by Jido library ?

lib/zaq/agent/tools/workflow/run_agent.ex
```diff
@@ -0,0 +1,129 @@
...
+
+        incoming = %Incoming{
+          content: resolved_input,
+          channel_id: "workflow:#{run_id || "anon"}",
+          author_id: "workflow",
+          provider: :workflow
+        }
+
+        outgoing =
+          executor.run(incoming,
```
jfayad: same comment as earlier, should go through a NodeRouter dispatch to benefit from all the pre-processing etc...

lib/zaq/agent/tools/workflow/dispatch_event.ex
jfayad, line 1: how is this action being used ? shouldn't we declare an action when a specific outcome is expected ?

This one is very abstract and can be misconfigured...

lib/zaq/agent/tools/workflow/iterate.ex
jfayad, line 1: is it the right abstraction ? should this live separately than the batch ? it seem to be the execution layer of the batch and seems like it should live there

lib/zaq/agent/tools/email/draft_reply.ex
jfayad: just saw that...

lib/zaq/engine/workflows/action.ex
jfayad, line 1: why didn't you use this as a superseed of a Jido.Action so all our tools actually `use` this one instead of `use Jido.Action` so validation can be made at compile time ?

I know this might come as a contradiction of my first comment on this file, but after seeing how tools declaration is done it's a nice inspiration and I would like to have your opinion on how feasible it would be

lib/zaq/engine/workflows/step_runner.ex
jfayad, line 1: it looks more like a runner than a wrapper....

lib/zaq/engine/workflows/cron_trigger_worker.ex
```diff
@@ -0,0 +1,52 @@
...
+  use Oban.Worker, queue: :default, max_attempts: 3
+
+  alias Zaq.Engine.Workflows
+  alias Zaq.Event
+
+  @impl Oban.Worker
+  def perform(%Oban.Job{args: %{"trigger_id" => trigger_id}}) do
+    case Workflows.get_trigger(trigger_id) do
+      nil ->
+        :ok
```
jfayad: you shouldn't just return ok, you should indicate a cancelation with the relevant reason

lib/zaq/engine/workflows/cron_trigger_worker.ex
```diff
@@ -0,0 +1,52 @@
...
+
+      %{enabled: false} ->
+        :ok
```
jfayad: same here, indicate why this job has been canceled

lib/zaq/engine/workflows/cron_trigger_worker.ex
```diff
@@ -0,0 +1,52 @@
...
+
+      %{trigger_type: type} when type != "cron" ->
+        :ok
```
jfayad: same comment here

lib/zaq/engine/workflows/startup_recovery.ex
jfayad, line 1: convert to Oban jobs so this can be properly tracked

jfayad (changes requested):
I have reviewed 55 files so far,

I left some to be technically demo'd

and others I still have to review indepedently

lib/zaq/engine/api.ex
```diff
@@ -311,7 +313,127 @@ defmodule Zaq.Engine.Api do
...
+
+  defp broadcast_run_update(run_id) do
+    case Workflows.get_run(run_id) do
+      %Zaq.Engine.Workflows.WorkflowRun{} = run ->
+        Phoenix.PubSub.broadcast(Zaq.PubSub, "workflow_run:#{run_id}", {:run_updated, run})
+
+      _ ->
+        :ok
+    end
+  end
```
jfayad: this doesn't belong to the API module which should be focused on gathering/shaping data and routing to the right module/function only

test/support/stubs.ex
jfayad, line 1: you are stubbing a Mock ? looks like a dirty workaround for a specific need you had at some point and that would require to be careful on how the mock is used anywhere else in tests..

priv/repo/migrations/20260520000001_create_workflow_approvals.exs
```diff
@@ -0,0 +1,26 @@
+defmodule Zaq.Repo.Migrations.CreateWorkflowApprovals do
```
jfayad: is it a workflow approval or a step approval ?

lib/zaq_web/live/bo/workflow_guard.ex
```diff
@@ -0,0 +1,17 @@
...
+  @moduledoc """
+  `on_mount` hook that blocks access to workflow LiveViews when the
+  workflows feature is disabled via the `WORKFLOWS_ENABLED` environment variable.
+  Redirects to the dashboard when the feature is off.
+  """
+  use ZaqWeb, :verified_routes
+  import Phoenix.LiveView
+
+  def on_mount(:require_workflows, _params, _session, socket) do
+    if Application.get_env(:zaq, :workflows_enabled, true) do
```
jfayad: this defaults to true, meaning it's enabled by default, it should return false by default

lib/zaq/engine/workflows/workflow_approval.ex
```diff
@@ -0,0 +1,62 @@
+defmodule Zaq.Engine.Workflows.WorkflowApproval do
```
jfayad: I think this should be renamed StepApproval

lib/zaq/agent/tools/workflow/batch.ex
```diff
@@ -0,0 +1,378 @@
...
+
+  ## How it works
+
+  `DagBuilder` resolves the `process` and `post_process` node lists and injects
+  them as `__batch_process__` and `__batch_post_process__` — lists of
+  `{module, base_params}` pairs.  It also inspects the input schema of the first
+  process action and injects `__batch_field__` (the key to deliver chunk data
+  under) and `__batch_mode__` (`:list` or `:item`).
+
+  Per chunk:
```
jfayad: chunking should support streams and lazily load chunks to feed them to the process

lib/zaq/agent/tools/workflow/batch.ex
```diff
@@ -0,0 +1,378 @@
...
+        type: {:list, :any},
+        required: true,
+        doc: "Injected by DagBuilder. List of {module, base_params} tuples for process pipeline."
+      ],
+      post_process: [
+        type: {:list, :any},
+        required: false,
+        default: [],
+        doc: "Injected by DagBuilder. Runs once per chunk after process completes."
+      ],
```
jfayad: you are reinventing the workflow in workflow logic, let's use it instead of redefining a different concept

lib/zaq/engine/workflows/dag_builder.ex
```diff
@@ -28,13 +28,67 @@ defmodule Zaq.Engine.Workflows.DagBuilder do
...
+              "module" => "Zaq.Engine.Workflows.Test.SleepMs",
+              "params" => %{"duration_ms" => 0}}
+          ]
+        },
+        "index" => 1
+      }
+
+  Inline nodes are validated against the `Workflows.Action` contract at build time.
+  String name references in `process`, `post_process`, or `pipeline` are rejected
+  with `{:error, :inline_node_required}`.
```
jfayad: all validation should happen when saving a workflow so we know the workflow representation we have in DB is valid and can be used immediately.

Moreover, each part of the workflow (trigger, step, edge etc...) should be in charge of its own validation logic. The main workflow validation should mostly delegate each part's validation and fail if any fail

lib/zaq/engine/workflows/dag_builder.ex
```diff
@@ -141,69 +196,244 @@ defmodule Zaq.Engine.Workflows.DagBuilder do
...
+        node
+        |> Map.put(:process, process)
+        |> Map.put(:post_process, post_process)
+        |> Map.put(:__batch_field__, field)
+        |> Map.put(:__batch_mode__, mode)
+        |> Map.update("params", %{}, &(&1 |> Map.delete("process") |> Map.delete("post_process")))
+
+      {:ok, enriched}
+    end
+  end
```
jfayad: shouldn't that belong to the batch node ?

lib/zaq/engine/workflows/dag_builder.ex
```diff
@@ -141,69 +196,244 @@ defmodule Zaq.Engine.Workflows.DagBuilder do
...
+      enriched =
+        node
+        |> Map.put(:__iterate_pipeline__, pipeline)
+        |> Map.put(:__iterate_field__, field)
+        |> Map.put(:__iterate_mode__, mode)
+        |> Map.update("params", %{}, &Map.delete(&1, "pipeline"))
+
+      {:ok, enriched}
+    end
+  end
```
jfayad: shouldn't that go to the iterate node ?

same for any node specific logic...

lib/zaq/engine/workflows/dag_builder.ex
```diff
@@ -141,69 +196,244 @@ defmodule Zaq.Engine.Workflows.DagBuilder do
...
+      {:cont, {:ok, [{mod, node_params} | acc]}}
+    else
+      {:error, _} = err -> {:halt, err}
+    end
+  end
+
+  defp first_action_batch_field([{first_mod, _} | _]), do: Action.batch_field(first_mod)
+  defp first_action_batch_field([]), do: {:error, {:missing_process_pipeline, :unknown}}
+
+  defp assemble(node_map, edges, run_id, batch_scoped) do
```
jfayad: I'm under the impression these functions that are building the dag should go into the run module to get the dag ready to be run

lib/zaq/engine/workflows/edge_condition.ex
```diff
@@ -0,0 +1,88 @@
...
+    do: raise(ArgumentError, "op :in requires a list expected value, got: #{inspect(expected)}")
+
+  def evaluate(op, _actual, _expected, _opts),
+    do: raise(ArgumentError, "unknown edge condition op: #{inspect(op)}")
+
+  defp empty?(nil), do: true
+  defp empty?(""), do: true
+  defp empty?([]), do: true
+  defp empty?(%{} = m) when map_size(m) == 0, do: true
+  defp empty?(_), do: false
```
jfayad: use changeset for that purpose, do not re-invent the wheel, plus it will be easier to add new conditions (or use existing contributions) down the road

lib/zaq/engine/workflows/workflow_agent.ex
```diff
@@ -49,40 +73,145 @@ defmodule Zaq.Engine.Workflows.WorkflowAgent do
...
+            Workflows.update_run(run, %{
+              status: "failed",
+              finished_at: DateTime.utc_now(:second),
+              log_summary: %{
+                error: format_build_error(reason),
+                failed_step_count: 0,
+                failed_steps: [],
+                step_count: 0
+              }
+            })
```
jfayad: this should not be a concern of the WorkflowAgent (which should probably be named WorkflowRunAgent).

It should receive a valid Dag in the WorkflowRun struct and run it

lib/zaq/engine/workflows.ex
```diff
@@ -115,15 +355,26 @@ defmodule Zaq.Engine.Workflows do
...
+        steps_snapshot: serialize_steps(workflow),
+        settings_snapshot: workflow.settings,
+        source_event: source_event,
+        status: "pending"
+      })
+      |> Repo.insert()
+
+    case result do
+      {:ok, run} ->
+        Phoenix.PubSub.broadcast(Zaq.PubSub, "workflow:#{run.workflow_id}", {:run_started, run})
```
jfayad: These should go through the NodeRouter.dispatch/1 to the Channels node

Moreover according to the logic you set, this doesn't belong here but inside the start_run/1 function

lib/zaq/engine/workflows.ex
```diff
@@ -205,32 +594,240 @@ defmodule Zaq.Engine.Workflows do
...
+  marks both as completed/approved in a transaction, transitions the run to `"paused"`,
+  then calls `resume_run/2`. The approval data flows to downstream steps via
+  ActionWrapper's resume idempotency cache.
+
+  Returns `{:error, :not_waiting}` if the run is not in `"waiting"` state.
+  Returns `{:error, :already_decided}` if the approval has already been acted on.
+  """
+  @spec approve_run(WorkflowRun.t(), WorkflowApproval.t(), map(), String.t() | nil, keyword()) ::
+          {:ok, WorkflowRun.t()} | {:error, :not_waiting | :already_decided | term()}
+  def approve_run(
```
jfayad: we are approving a step, not a full run

lib/zaq/engine/workflows.ex
```diff
@@ -205,32 +594,240 @@ defmodule Zaq.Engine.Workflows do
...
+
+  @doc """
+  Rejects a waiting workflow run, failing the approval step and the run.
+
+  Returns `{:error, :not_waiting}` if the run is not in `"waiting"` state.
+  Returns `{:error, :already_decided}` if the approval has already been acted on.
+  """
+  @spec reject_run(WorkflowRun.t(), WorkflowApproval.t(), String.t(), String.t() | nil, keyword()) ::
+          {:ok, WorkflowRun.t()} | {:error, :not_waiting | :already_decided | term()}
+  def reject_run(
```
jfayad: we reject a step, not a full run

lib/zaq/engine/workflows.ex
```diff
@@ -448,4 +1164,97 @@ defmodule Zaq.Engine.Workflows do
...
+  """
+  def broadcast_batch_progress(nil, _step_name, _progress), do: :ok
+
+  def broadcast_batch_progress(run_id, step_name, progress) do
+    Phoenix.PubSub.broadcast(
+      Zaq.PubSub,
+      "workflow_run:#{run_id}",
+      {:batch_progress, step_name, progress}
+    )
+  end
```
jfayad: shoulf use the NodeRouter.dispatch/1 in async mode

Same for all broadcasts

jfayad (changes requested):

lib/zaq/engine/workflows.ex
jat10: make sure that the start_run and create_run can run separately or they should together

config/runtime.exs
jat10: http1 is common between providers 
Also, for the real management of this problem, I created this issue #493

jfayad: common between providers is not the question, why not using http2 ?

Also the issue is about throttling not specifically about the protocol used to issue requests

lib/zaq/engine/workflows/action.ex
```diff
@@ -3,33 +3,32 @@ defmodule Zaq.Engine.Workflows.Action do
...
-  *additionally* declares this behaviour:
+  A workflow action is a `Jido.Action` (for `run/2`, parameter validation, and
+  the `schema` / `output_schema` introspection) that *additionally* declares
+  this behaviour. The idiomatic way is a single `use` call — this macro wraps
+  `use Jido.Action`, attaches the behaviour with default `on_success/2` /
+  `on_failure/2`, and enforces the contract at compile time:
 
       defmodule MyAction do
-        use Jido.Action,
+        use Zaq.Engine.Workflows.Action,
```
jfayad, line 13: this approach forces to redefine all actions to use this macro.

Why not keep the action code as a standard Jido Action and provide a wrapper module that can offer an "adapter" to the actions for them to be usable in workflow: similar to how jido_ai exposes actions as tools without having to redefine them

lib/zaq/agent/tools/workflow/dispatch_event.ex
```diff
@@ -0,0 +1,113 @@
...
+        type: :string,
+        required: true,
+        doc: "Allowlisted Engine event name, e.g. \"lead_identified\""
+      ],
+      machine: [
+        type: :boolean,
+        required: false,
+        default: false,
+        doc:
+          "Mark the dispatched event as a machine (actorless) run, setting skip_permissions on the triggered run."
```
jfayad, line 90: I'm not sure this is the right approach here (to consider a machin run has all permissions), instead machine runs should carry on an explicit role/user to run on behalf of and to carry the permissions for

lib/zaq/agent/tools/accounts/history.ex
```diff
@@ -0,0 +1,231 @@
...
+        type: :integer,
+        required: false,
+        default: 50,
+        doc: "Maximum number of messages to include per conversation."
+      ],
+      person_id: [
+        type: :any,
+        required: false,
+        doc:
+          "Target person ID. Honored only on machine/admin runs (skip_permissions); otherwise ignored — identity comes from the execution context."
```
jfayad, line 87: this whole concept of `skip_permissions` looks like a bad idea... it's a gate to hard to debug permission breaches

lib/zaq/agent/tools/sheets/update_sheet_values.ex
```diff
@@ -54,7 +75,17 @@ defmodule Zaq.Agent.Tools.UpdateSheetValues do
       :data_source_sheet_update_values,
       request,
       context,
-      "Data source sheet update failed"
+      "Data source sheet update failed",
+      &validate_sheet_response/1
     )
   end
+
+  defp validate_sheet_response(%{record: %Record{}} = payload), do: {:ok, payload}
```
jfayad, line 83: these are being copied accross multiple sheet tools, why not centralizing these into the DataSourceTools module?

lib/zaq/agent/tools/workflow/dispatch_event.ex
jat10: this is made like this now but later on we will support resource management where each workflow have it's own permissions.

lib/zaq/agent/tools/accounts/history.ex
jat10: this is a mechanisim that we created before for bo users and it's applied here as well. So this should be handled in channels when it will be used from mattermost, ... that skip_permission will not be there

lib/zaq/agent/tools/workflow/dispatch_event.ex
jfayad: it is not about defining a permission on the workflow level: I think this is also a bad idea.

it is about enforcing the permission based on the user/role the workflow is running as... workflows do not carry permissions by themselves, the workflow runner do

jat10: But @jfayad this case is supported because we have a workflow that is dispatching an event to trigger another workflow. And these could be run using a cron so how you will handle the user permissions which user ran them?

lib/zaq/agent/tools/pipeline_runner.ex
jat10: This file was removed it's was created for batch and itteration inline processing

lib/zaq/engine/workflows/startup_recovery.ex
jat10: `RunRecoveryWorker` is an Oban job

jfayad: Then why does it need to be running in a Supervised Genserver ?

If that's on application start, it should be reflected in the code

lib/zaq/agent/tools/workflow/dispatch_event.ex
jfayad: Ideas:

- When invoked by another workflow run: use the same caller by default (could be superseeded when building the workflow)
- When invoked by a cron: define the user the cron should run as

lib/zaq/agent/tools/accounts/history.ex
jfayad: The context matters here... when we are in the BO the intent is clear.

When we are in a workflow run, this concept should not exist...

config/runtime.exs
jat10: No deliberate decision `:http1` is ReqLLM's default, and I kept it

jfayad: keeping http1 is safer and here is the reason

>ReqLLM will error with a helpful message if you try to send a large request body with mixed HTTP/1+HTTP/2 pools.

lib/zaq/agent/tools/sheets/update_sheet_values.ex
jat10: This is being added to #502

lib/zaq/engine/api.ex
jat10: this being added to #502

lib/zaq/engine/workflows/action.ex
jat10: This being added to #502

lib/zaq/agent/tools/workflow/dispatch_event.ex
jat10: #504

lib/zaq/agent/tools/accounts/history.ex
jat10: #504

lib/zaq/engine/workflows/dag_builder.ex
jat10: #502

lib/zaq/engine/workflows/dag_builder.ex
jat10: #506

lib/zaq/agent/api.ex
```diff
@@ -359,6 +368,7 @@ defmodule Zaq.Agent.Api do
             team_ids: team_ids,
             source_filter: incoming.content_filter,
             skip_permissions: Keyword.get(pipeline_opts, :skip_permissions, false),
+            system_prompt: Keyword.get(pipeline_opts, :system_prompt),
```
jfayad: this shouldn't be possible...

lib/zaq/agent.ex
```diff
@@ -64,6 +64,14 @@ defmodule Zaq.Agent do
...
   end
 
+  @spec get_agent_by_name(String.t()) :: {:ok, ConfiguredAgent.t()} | {:error, :agent_not_found}
+  def get_agent_by_name(name) when is_binary(name) do
+    case Repo.get_by(ConfiguredAgent, name: name) do
+      nil -> {:error, :agent_not_found}
+      agent -> {:ok, agent}
+    end
+  end
+
```
jfayad, line 74: is this being used ?

jfayad: it's not but we can keep it if needed later on

lib/zaq/engine/workflows/startup_recovery.ex
jfayad: #511

jfayad (approved):

PR #431, opened 2026-05-27
feat: add storybook

jana914:
add storybook to review components and define standardized design system


.claude/settings.json
jfayad, line 1: I'm not sure you should push this file, @jat10 what's your position on that ?

config/config.exs
```diff
@@ -112,7 +112,7 @@ config :esbuild,
     env: %{"NODE_PATH" => [Path.expand("../deps", __DIR__), Mix.Project.build_path()]}
   ]
 
-config :license_manager, :repo, Zaq.Repo
+# config :license_manager, :repo, Zaq.Repo
```
jfayad: @jat10 what do you suggest doing on that level ?

config/dev.exs
```diff
@@ -3,10 +3,11 @@ import Config
...
   types: Zaq.PostgrexTypes,
-  username: "postgres",
-  password: "postgres",
-  hostname: "localhost",
-  database: "zaq_dev",
+  username: System.get_env("DB_USER", "postgres"),
+  password: System.get_env("DB_PASSWORD", "postgres"),
+  hostname: System.get_env("DB_HOST", "localhost"),
+  database: System.get_env("DB_NAME", "zaq_dev"),
+  port: 5431,
```
jfayad: why did you need to put a custom port ? 5432 is the default one...

If you can't use the default port, better to set your custom port in your .env and leave 5432 by default

lib/zaq_web/components/layouts/root.html.heex
```diff
@@ -8,6 +8,8 @@
     <.live_title default="ZAQ" suffix=" · Back Office">
       {assigns[:page_title]}
     </.live_title>
+    <script src="https://mcp.figma.com/mcp/html-to-design/capture.js" async>
+    </script>
```
jfayad: We certainly don't want to leak this file to the global layout, don't you have a way to get this dynamically injected in dev only ?

test/storybook/package.json
jfayad, line 1: we already have an e2e test suite, let's unify the storybook tests with the existing one to avoid drift with playwright deps

jfayad (changes requested):

mix.exs
```diff
@@ -65,6 +65,7 @@ defmodule Zaq.MixProject do
       {:postgrex, ">= 0.0.0"},
       {:phoenix_html, "~> 4.1"},
       {:phoenix_live_reload, "~> 1.2", only: :dev},
+      {:phoenix_storybook, "~> 0.8", only: :dev},
```
jfayad: you are declaring this module for dev. only but you did put some code files inside the folder where it gets compiled for other environment too (test, prod)

This is failing compilation in CI jobs

.claude/settings.json
jat10: `settings.json` is intended to be shared at the project level, so everyone using the same Claude setup has access to the same functions and configuration.

For personal MCPs and permissions, use `settings.local.json` instead. This file is gitignored and meant for user-specific customizations.

Please revert the changes made to `settings.json` and move them to `settings.local.json`.

config/config.exs
jat10: Let us keep it that way for now. I will handle that later.

config/dev.exs
jat10: I suggest editing that in dev.secret.exs

config/config.exs
jfayad: when is later ? we can't have no license_manager in a release...

jana914: What do you mean, keep it that way ?

jat10: @jana914, no need to do anything in the file.

jfayad (approved):

PR #468, opened 2026-06-17
fix(ingestion): preserve document headings by storing chunk title separately

### LINKED ISSUE #463: [ingestion]: Loosing data while chunking files
What is happening now is that when we are ingesting a file, the document chunker is handling the chunking, but the title generator is replacing the header with a semantic data enrichment, which is leading to losing the original heading data.

jat10:
fix #463


jfayad:
it doesn't look like this new title is being used anywhere... Can you explain why do we have to compute it and store it ?

PR #505, opened 2026-06-24
feat(agent): add single-cell sheet updates and integer-coercing increment tool

### LINKED ISSUE #509: [Actions] Add metadata to accounts history
In order to enable more workflow use cases we would need to add a `metadata` field to the `Zaq.Agent.Tools.Accounts.History` action

this object will have the following shape:

```json
"metadata": {
  "total": {
    "messages": 3,
    "user_messages": 1,
    "assistant_messages": 2,
    "first_message_date": {date_or_timestamp},
    "last_message_date": {date_or_timestamp}
  },
  "current_window": {
    "messages": 1,
    "user_messages": 0,
    "assistant_messages": 1,
    "first_message_date": {date_or_timestamp},
    "last_message_date": {date_or_timestamp}
  }
}
```

`total` will be for the whole conversation regardless of what was loaded after a filtered query
`current_window` will be computed after the filters have been applied

jat10:
Fixes #509


lib/zaq/agent/tools/sheets/update_sheet_values.ex
jfayad, line 1: Keep actions code as simple as possible and do not leak specific concerns into them (here the need to provide a targeted cell as separate variables instead of a valid range string)

lib/zaq/agent/tools/data_source/create_folder.ex
jfayad, line 1: if creating a folder is just another file with a specific mime type, then this action should not exist and we should simply use the create_file action with the correct mime type.

Moreover, you are having a very narrow view of the action... this is a data source action not a google drive specific one, so no mention of google drive specifics (i.e the `@folder_mime_type "application/vnd.google-apps.folder"` should not be here)

lib/zaq/agent/tools/workflow/increment.ex
jfayad, line 1: there is an existing action to increment, use it instead of declaring a new one

lib/zaq/agent/tools/workflow/run_agent.ex
```diff
@@ -57,7 +61,9 @@ defmodule Zaq.Agent.Tools.Workflow.RunAgent do
   defp dispatch_agent_run(agent, input, params, context) do
     node_router = Map.get(context, :node_router, NodeRouter)
     vars = build_vars(params)
-    system_prompt = substitute(agent.job || "", vars)
+    # System prompt is kept verbatim (static) so the prompt prefix stays
+    # cacheable across runs; only the user message carries run-specific data.
+    system_prompt = agent.job || ""
```
jfayad: you don't need to pass it to the call... it gets picked up from the agent when not provided

lib/zaq/agent/tools/workflow/run_agent.ex
jfayad: you don't need to redeclare an agent verification logic here... just dispatch the call to the agent and let the agent node handle the failures there, this also means no need to load the agent here...

lib/zaq/agent/tools/workflow/run_agent.ex
```diff
@@ -2,17 +2,21 @@ defmodule Zaq.Agent.Tools.Workflow.RunAgent do
...
-  Looks up the agent by `agent_name` from the database, retrieves its `job`
-  field (system prompt), substitutes `{{variable}}` placeholders using all
-  accumulated workflow params, then invokes `Executor.run/2`.
+  Looks up the agent by `agent_name` from the database, uses its `job` field
+  (system prompt) verbatim, substitutes `{{variable}}` placeholders in `input`
+  using all accumulated workflow params, then invokes `Executor.run/2`.
+
+  The system prompt is deliberately **not** templated: keeping it static lets
+  providers cache the prompt prefix across runs. All run-specific/custom data
+  belongs in `input` (the user message), never in the system prompt.
```
jfayad: - not necessary to lookup the agent from the DB, having the agent name is all we need to dispatch the event to the Agent node
- not necessary to comment on what is not done (e.g system prompt), especially when it's about stating the obvious

lib/zaq/agent/tools/registry.ex
```diff
@@ -64,6 +64,12 @@ defmodule Zaq.Agent.Tools.Registry do
       description: "Create a document on a specific datasource provider",
       module: Zaq.Agent.Tools.DataSource.CreateDocument
     },
+    %{
+      key: "data_source.create_folder",
+      label: "Create folder",
+      description: "Create a folder on a specific datasource provider",
+      module: Zaq.Agent.Tools.DataSource.CreateFolder
+    },
```
jfayad: to remove

jfayad (changes requested):

lib/zaq/agent/tools/workflow/increment.ex
jat10: this is an issue we created for now we need to be declared as `use Zaq.Engine.Workflows.Action` to be used inside a workflow

lib/zaq/agent/tools/sheets/update_sheet_values.ex
jat10: fixed it and i added a concat action to support our use case

lib/zaq/agent/tools/workflow/increment.ex
jfayad: I know we created an issue, but in the meantime don't re-implement, use delegate instead

lib/zaq/agent/tools/workflow/run_agent.ex
jfayad: I saw the changes you did but this is all wrong...

The agent node code should NOT be impacted by this action!

This action should expect to have an agent identifier that is similar to how channels are pointing to agents so the exact same calls can be leveraged

lib/zaq/agent/tools/workflow/run_agent.ex
```diff
@@ -41,23 +43,18 @@ defmodule Zaq.Agent.Tools.Workflow.RunAgent do
 
   require Logger
 
-  alias Zaq.Agent
   alias Zaq.Engine.Messages.{Incoming, Outgoing}
   alias Zaq.Event
   alias Zaq.NodeRouter
 
   @impl Jido.Action
   def run(%{agent_name: agent_name, input: input} = params, context) do
```
jfayad: If the Agent's node requires the agent id when being called, store the id instead of the name...

Double check how it's done in channels and use the same approach so you can re-use the same calls

lib/zaq/agent/api.ex
jfayad, line 1: this should not be edited at all, if anything you should remove the changes you pushed in the main workflow PR to support system prompts override, but that's all

lib/zaq/agent/executor.ex
jfayad, line 1: This file should not be touched at all!

lib/zaq/agent.ex
```diff
@@ -93,6 +93,15 @@ defmodule Zaq.Agent do
...
 
+  @spec get_active_agent_by_name(String.t()) :: {:ok, ConfiguredAgent.t()} | {:error, atom()}
+  def get_active_agent_by_name(name) when is_binary(name) do
+    case get_agent_by_name(name) do
+      {:ok, %ConfiguredAgent{active: true} = agent} -> {:ok, agent}
+      {:ok, %ConfiguredAgent{}} -> {:error, :inactive_agent}
+      {:error, :agent_not_found} -> {:error, :agent_not_found}
+    end
+  end
+
```
jfayad: these two functions have been added for Workflow concerns... are these valid concerns or should these be removed ?

jfayad (changes requested):

lib/zaq/agent/tools/data_source/create_document.ex
```diff
@@ -17,17 +17,21 @@ defmodule Zaq.Agent.Tools.DataSource.CreateDocument do
...
       name: [type: :string, required: true, doc: "Document name/title"],
-      content: [type: :string, required: true, doc: "Textual content to create"],
+      content: [type: :string, required: false, doc: "Optional textual content to create"],
       path: [type: :string, required: false, doc: "Optional provider path/parent folder"],
       parent_id: [type: :string, required: false, doc: "Optional provider parent identifier"],
+      parents: [
+        type: {:list, :string},
+        required: false,
+        doc: "Optional provider parent identifiers"
+      ],
```
jfayad: this seem redundant with the parent_id ?

lib/zaq/agent/tools/workflow/increment.ex
jfayad: @jat10 you have missed this one

jfayad (approved):

PR #522, opened 2026-06-26
feat(channels): route incoming messages to agent or simply fire an event with no-op - [channels]: trigger a workflow

### LINKED ISSUE #503: [channels]: trigger a workflow
Image

We need to allow users to set a workflow rather an agent only. So a trigger should be created and triggered when a message got received by a channel to run the workflow

jfayad:
Fixes #503

### DIFF
--- a/lib/zaq/agent/pipeline.ex
+++ b/lib/zaq/agent/pipeline.ex
@@ -46,6 +46,7 @@ defmodule Zaq.Agent.Pipeline do
   alias Zaq.Engine.Messages.{Incoming, Outgoing}
   alias Zaq.Engine.Telemetry
   alias Zaq.Event
+  alias Zaq.Identity.ActorNormalizer
 
   # ---------------------------------------------------------------------------
   # Public API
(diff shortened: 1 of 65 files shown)


lib/zaq/engine/messages/incoming.ex
```diff
@@ -123,6 +136,30 @@ defmodule Zaq.Engine.Messages.Incoming do
...
+    end
+  end
+
+  defp normalize_person(_), do: nil
+
+  defp person_field(person, key) when is_map(person) and is_atom(key) do
+    Map.get(person, key) || Map.get(person, Atom.to_string(key))
+  end
+
+  defp person_field(_person, _key), do: nil
```
jat10: These are defined in `Zaq.Identity.ActorNormalizer` let us centralize the logic

lib/zaq/engine/messages/outgoing.ex
```diff
@@ -56,7 +56,10 @@ defmodule Zaq.Engine.Messages.Outgoing do
       author_name: incoming.author_name,
       provider: incoming.provider,
       in_reply_to: incoming.message_id,
-      metadata: incoming.metadata |> Map.merge(result) |> Map.put(:person_id, incoming.person_id)
+      metadata:
+        incoming.metadata
+        |> Map.merge(result)
+        |> Map.put(:person_id, Incoming.person_id(incoming))
```
jat10: If the incoming has the person, where we need only to have the `person_id` let us unify the experience and have the person object

lib/zaq/agent/pipeline.ex
```diff
@@ -54,32 +55,26 @@ defmodule Zaq.Agent.Pipeline do
...
-           ).response do
-        nil -> []
-        %{team_ids: ids} when not is_nil(ids) -> ids
-        _ -> []
-      end
+
+    actor =
+      opts |> Keyword.get(:event) |> event_actor() || ActorNormalizer.from_incoming(nil, incoming)
+
+    person_id = Keyword.get(opts, :person_id) || ActorNormalizer.person_id(actor)
```
jat10, line 62: why we still need person_id here

jat10 (changes requested):

lib/zaq/engine/messages/incoming.ex
jfayad: done

lib/zaq/engine/messages/outgoing.ex
jfayad: done

lib/zaq/agent/pipeline.ex
jfayad: it is used for permission related verification, not worth refactoring this now as it should be broken down into actions and converted to a workflow

jat10 (approved):

PR #599, opened 2026-07-16
fix(ingestion): chunks preserve data 

jat10:


lib/zaq/ingestion/document_chunker.ex
```diff
@@ -36,16 +36,62 @@ defmodule Zaq.Ingestion.DocumentChunker do
...
+    @moduledoc """
+    Represents a chunk with metadata.
+
+    Invariants:
+
+      * `content` is a byte-exact substring of the source document — never
+        synthesized, trimmed, or rewritten.
+      * `embedding_input` is the embed-only enrichment (section-path context
+        prefix + `content`). It is transient: sent to the embedding model,
+        never persisted, FTS-indexed, or shown to users.
```
jfayad: why do we need a separate input for the embedding ? I would expect the embedding to be based on the content...

jfayad (approved):

PR #612, opened 2026-07-22
feat(channels): deliver telegram messages as rich markdown for table support

jat10:


lib/zaq/channels/jido_chat_bridge.ex
```diff
@@ -38,6 +38,11 @@ defmodule Zaq.Channels.JidoChatBridge do
   alias Zaq.Channels.JidoChatBridge.State
   alias Zaq.Engine.Messages.{Incoming, Outgoing}
   import Zaq.Engine.Messages, only: [is_present_message_id: 1]
+
+  # Formats an adapter may receive on the wire. `:rich_markdown` / `:rich_html`
+  # ship the body unconverted so the platform parses it server-side — that is how
+  # Telegram renders tables, which its `sendMessage` HTML tag set cannot express.
+  @deliverable_formats [:html, :plain_text, :markdown, :rich_markdown, :rich_html]
```
jfayad: these formats should be standardized and not leak some adapter's specific formats.

IMO, the `:markdown` format is what should be sent to Telegram and it should be transformed to `:rich_markdown` on the jido_chat_telegram code...

My understanding is that `markdown` on telegram side is a subset of markdown and `rich_markdown` is just markdown

config/config.exs
```diff
@@ -25,8 +25,7 @@ config :zaq, :channels, %{
     bridge: Zaq.Channels.JidoChatBridge,
     adapter: Jido.Chat.Telegram.Adapter,
     ingress_mode: :webhook,
-    message_format: :html,
-    message_formatter: {Telegex.Marked, :as_html}
+    message_format: :rich_markdown
```
jfayad: if you agree with the logic on my first comment, the only change in this PR is the removal of telegex and setting the message_format to `markdown` (we might even just remove it from the config as I think the default format is already `markdown`

jfayad (changes requested):

lib/zaq/channels/message_formatter.ex
jfayad, line 1: the changes in this file are useless

lib/zaq/channels/jido_chat_bridge.ex
jfayad, line 1: the changes in this file are useless

config/config.exs
```diff
@@ -25,8 +25,7 @@ config :zaq, :channels, %{
     bridge: Zaq.Channels.JidoChatBridge,
     adapter: Jido.Chat.Telegram.Adapter,
     ingress_mode: :webhook,
-    message_format: :html,
-    message_formatter: {Telegex.Marked, :as_html}
+    message_format: :markdown
```
jfayad: markdown is the default format, this line can be removed

docs/services/channels.md
```diff
@@ -275,8 +275,7 @@ config :zaq, :channels, %{
   telegram: %{
     bridge: Zaq.Channels.JidoChatBridge,
     adapter: Jido.Chat.Telegram.Adapter,
-    message_format: :html,
-    message_formatter: {Telegex.Marked, :as_html}
+    message_format: :markdown
```
jfayad: this was an example on how to use a message_formatter, I would keep it and precise it's an example

jfayad (changes requested):

jfayad (approved):

PR #620, opened 2026-07-23
feat: bo create file - LLM creates files via DiskBridge

Hassanhanyq:


lib/zaq/agent/tools/files/create_file.ex
```diff
@@ -0,0 +1,57 @@
...
+      else
+        "generated/#{out_name}"
+      end
+
+    {:ok,
+     %{
+       name: out_name,
+       path: rel_path,
+       mime_type: out_mime,
+       url: "/bo/files/#{rel_path}",
```
jat10: why? hard coded this is wrong

lib/zaq/agent/tools/files/persist_file.ex
```diff
@@ -0,0 +1,79 @@
...
+          |> Map.put(:content, Base.encode64(params.data))
+          |> Map.drop([:data])
+
+        event =
+          encoded_params
+          |> Event.new(:channels, opts: [action: :disk_persist_file])
+          |> NodeRouter.dispatch()
+
+        Map.fetch!(event, :response)
+    end
```
jat10: This logic should be handled by the bridge. The only pre-dispatch responsibility here is to default to `disk` when no provider is specified.

test/zaq/agent/tools/files/create_file_test.exs
```diff
@@ -0,0 +1,55 @@
+defmodule Zaq.Agent.Tools.Files.CreateFileTest do
+  use ExUnit.Case, async: true
+
+  alias Zaq.Agent.Tools.Files.CreateFile
+
+  @ctx %{}
+
```
jat10: there is no coverage for Create.run with missing required fields

lib/zaq/agent/tools/files/persist_file.ex
```diff
@@ -0,0 +1,79 @@
+defmodule Zaq.Agent.Tools.Files.PersistFile do
```
jat10: There is no persistFile test coverage

jat10 (changes requested):

lib/zaq/channels/api.ex
```diff
@@ -294,6 +294,47 @@ defmodule Zaq.Channels.Api do
     %{event | response: data_source_module.create_file(provider, params)}
   end
 
+  def handle_event(
```
jat10: @Hassanhanyq so now this function is never called right?

jat10 (changes requested):

lib/zaq/agent/tools/general/build_http_request.ex
jfayad, line 1: since you are building a `Req` request and already take a timeout input, why do we need a separate execute http request action ?

We should probably dispatch the request in this action so the agent will get the actual result in one turn

lib/zaq/agent/tools/general/list_providers.ex
```diff
@@ -0,0 +1,75 @@
...
+    ],
+    output_schema: [
+      providers: [
+        type: {:list, :map},
+        required: true,
+        doc: ~s|Provider menu — each entry has a `provider` key and a `label`|
+      ]
+    ]
+
+  alias Zaq.Agent.Tools.DataSourceTool
```
jfayad: should that move to the general folder ?

lib/zaq/agent/tools/general/list_providers.ex
```diff
@@ -0,0 +1,75 @@
...
+        doc:
+          ~s|Channel kind to list providers for: "data_source" for document | <>
+            ~s|destinations, "communication" for message destinations|
+      ]
+    ],
+    output_schema: [
+      providers: [
+        type: {:list, :map},
+        required: true,
+        doc: ~s|Provider menu — each entry has a `provider` key and a `label`|
```
jfayad: let's the return also include the status of the channel so it can know if the channel has an active configuration or not

lib/zaq/channels/req.ex
```diff
@@ -0,0 +1,332 @@
...
+
+  ## Secret placeholders are not resolved yet
+
+  `BuildHttpRequest` renders credentials as `{{secret:name}}` placeholders.
+  **This module does not substitute them** — it sends the spec exactly as
+  given, so a request carrying a placeholder goes out with the literal
+  `{{secret:name}}` text and the remote API will reject it. Wiring resolution
+  to a credential store is deliberately deferred; until then this executor is
+  only useful for endpoints that need no authorization, or for callers that
+  substitute values into the spec themselves before calling.
```
jfayad: what's the point of introducing this type of security at the cost of a non functional feature ?

lib/zaq/channels/req.ex
```diff
@@ -0,0 +1,332 @@
...
+  defp normalize_url(url), do: {:error, {:invalid_url, url}}
+
+  # -- Destination safety -----------------------------------------------------
+
+  defp authorize_destination(url, config) do
+    with {:ok, %URI{scheme: scheme, host: host}} <- parse_uri(url),
+         :ok <- check_scheme(scheme, url),
+         :ok <- check_host_present(host, url),
+         :ok <- check_allowed_hosts(host, Keyword.get(config, :allowed_hosts, [])) do
+      check_address(host, config)
```
jfayad: all these checks should happen when building the http request, I see no reason why these shall be happening when executing the request...

lib/zaq/channels/req.ex
```diff
@@ -0,0 +1,332 @@
...
+       %{
+         method: method,
+         url: url,
+         headers: get(spec, :headers) || %{},
+         query: get(spec, :query) || %{},
+         body: get(spec, :body),
+         body_format: get(spec, :body_format) || "json"
+       }}
+    end
+  end
```
jfayad: what is this ? again all of these should be done at build time...

lib/zaq/channels/communication_bridge.ex
jfayad, line 1: these are now duplicated in communication and data_source bridges... looks like it should go into the global bridge module

lib/zaq/channels/api.ex
```diff
@@ -623,6 +630,17 @@ defmodule Zaq.Channels.Api do
     end
   end
 
+  def handle_event(%Event{request: %{request: request}} = event, :http_request, _context)
+      when is_map(request) do
+    # Full name on purpose: a bare `Req` alias here would shadow the external
+    # Req library and read as the wrong module.
+    http_module = Keyword.get(event.opts, :http_module, Zaq.Channels.Req)
+    %{event | response: http_module.request(request, event.opts)}
```
jfayad: is that needed can't we just use the Req module ?

jfayad (changes requested):

lib/zaq/agent/tools/general/list_providers.ex
jat10: You mean this tool to be more general like Zaq.Agent.Tools.Channels and do the dispatch to datasource?

jfayad: I mean you moved the action list_providers to the general folder because it's not for DataSource only.

The helper should also be adjusted as it points to all Channels not only DataSource ones

Just a naming/code organization for consistency

jat10:
The following PR should be teared down into 5 small PRs:
- Http request action 
- Decoding and encoding action
- Convertor action 
- Ability to receive an image from req_llm and jido_ai that is blocked by 2 issues
- Disk bridge and create document

PR #623, opened 2026-07-24
feat(ingestion): add md_to_doc python step and fetch it in zaq.python fetch

jat10:


jfayad:
Rejected: new features should not resort to hacks.

PR #627, opened 2026-07-25
feat(channels): allow user to react on message through communication channels

jat10:


lib/zaq/channels/jido_chat_bridge/reaction_payload.ex
```diff
@@ -0,0 +1,183 @@
...
+
+  Listener sinks deliver reactions in provider-specific shapes: Discord's gateway
+  worker sends a `%Jido.Chat.EventEnvelope{}` struct, other transports send plain
+  maps with atom or string keys. This module collapses all of them into the single
+  normalized struct that `JidoChatBridge.handle_reaction_event/2` consumes, so the
+  in-process `Jido.Chat.on_reaction/2` handler and the listener sink path feed it
+  identical data.
+
+  Normalization is total: an unusable payload returns `:ignored` rather than
+  raising, because it runs inside the bridge state process.
```
jfayad: looks like this module is trying to re-invent the wheel... these ReactionEvent structs should be built at the boundaries (in the Chat adaptors) not in here...

lib/zaq/channels/jido_chat_bridge.ex
```diff
@@ -326,6 +333,53 @@ defmodule Zaq.Channels.JidoChatBridge do
...
+        {:ok, %{follow_up_text: text}} when is_binary(text) ->
+          post_reaction_follow_up(config, reaction.thread, text)
+
+        _ ->
+          :ok
+      end
+    else
+      _ -> :ok
+    end
+  end
```
jfayad, line 369: You should look at how this is done for handling message events and do the same here: let the global CommunicationBridge take care of that

lib/zaq/engine/api.ex
```diff
@@ -514,6 +514,19 @@ defmodule Zaq.Engine.Api do
     end
   end
 
+  def handle_event(%Event{} = event, :rate_message_from_reaction, _context) do
```
jfayad: why is that specific to reactions ? we are mapping reactions to ratings in the channels' node, this should then be processed by the same path for BO based ratings, not go through a different path

lib/zaq/engine/conversations.ex
```diff
@@ -684,6 +695,35 @@ defmodule Zaq.Engine.Conversations do
...
+      #   :ok
+      # end
+      :ok
+    else
+      nil -> :ok
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  def rate_message_from_reaction(_), do: :ok
```
jfayad: Same remark as for the API, this shouldn't be specific to reactions based ratings... it should go through the same code path...

Channels is in charge of mapping reactions to valid ZAQ ratings, then the rating should flow inside ZAQ's node as a standardize payload regardless of its origin

lib/zaq/channels/jido_chat_bridge/state.ex
```diff
@@ -313,4 +319,17 @@ defmodule Zaq.Channels.JidoChatBridge.State do
   defp chat_module do
     bridge_module().runtime_chat_module()
   end
+
+  defp payload_event(payload), do: ReactionPayload.event_type(payload)
```
jfayad: this is NOT a Reaction specific helper, it's a global message one

lib/zaq/channels/jido_chat_bridge/state.ex
```diff
@@ -313,4 +319,17 @@ defmodule Zaq.Channels.JidoChatBridge.State do
...
+
+  defp handle_reaction_payload(config, payload) do
+    provider = JidoChatBridge.provider_to_atom(config.provider) || :unknown
+
+    with {:ok, adapter} <- bridge_module().adapter_for(config.provider),
+         {:ok, reaction} <- ReactionPayload.to_reaction_event(payload, provider, adapter) do
```
jfayad: why are you redefining this logic ? isn't it the role of `Chat.process_reaction/4` ? (Analog to `Chat.process_message/5`)

jfayad (changes requested):

lib/zaq/engine/api.ex
```diff
@@ -514,10 +514,54 @@ defmodule Zaq.Engine.Api do
     end
   end
 
+  def handle_event(%Event{} = event, :rate_message, _context) do
```
jfayad, line 517: should you have a follow up issue to make sure the BO rating is also using this same event?

jat10: created an issue for that #630

jfayad (dismissed):

PR #644, opened 2026-07-30
feat(live): add skill resource uploads backed by ingestion volumes

### LINKED ISSUE #645: [agent]: Support file management for skill references on the BO
# Summary

To let agent skills use resources, we need to support file management at the BO (back office) level. This first pass focuses on `references`, which are saved to disk at:

```
.agents/skills/{skill-name}/references/{file-name}
```

## Requirements

- Required fields: `name`, `description`, and `instructions`.
- No volume linked: if the skill has no volume linked, show a message prompting the user to add one.
- Volume switching: while uploading a file, the user can switch between different volumes.
- Temp upload handling: since the destination is keyed by `skill-name`, and the skill may not exist yet at upload time, files are staged using Phoenix LiveView's upload buffer. This lets us move the file to the correct destination once the skill is saved even if the user changes the skill name before saving. If the user cancels without creating the skill, Phoenix handles cleanup of the temp file automatically.
- Deletion: deleting a skill that has an associated resource deletes its directory, removing `{skill-name}/references/{file-name}`.

jat10:
fixes #645 

Here is a screenshot of how Resources is embedded inside the Skill drawer:
Screenshot 2026-07-30 at 3 56 43 PM


lib/zaq_web/components/design_system/modal_no_volume.ex
```diff
@@ -0,0 +1,55 @@
+defmodule ZaqWeb.Components.DesignSystem.ModalNoVolume do
```
jfayad: why do you need to create a new component just for that ?
You should re-use the existing components

lib/zaq_web/components/design_system/modal_skill_resources.ex
```diff
@@ -0,0 +1,81 @@
+defmodule ZaqWeb.Components.DesignSystem.ModalSkillResources do
```
jfayad: why do you need a new component just for that ? can't you re-use the same one used to upload a file in the ingestion page ?

lib/zaq_web/helpers/upload_flash.ex
```diff
@@ -0,0 +1,53 @@
+defmodule ZaqWeb.Helpers.UploadFlash do
```
jfayad: make sure to rebase and check if the changes made in the UI bring in new components that you can use for that purpose (especially in the ingestion flash confirmation)

jfayad (changes requested):
Check the questions in the issue which would probably impact the code you pushed already and is outside of the current code reviews

lib/zaq_web/components/design_system/modal_skill_resources.ex
jat10: # Upload component on ingestion
Screenshot 2026-07-30 at 3 43 41 PM

# Upload on resources

Since we are allowing the user to pick the volume they want there is a change in the component. We are using some same drop zone component 
Screenshot 2026-07-30 at 3 43 29 PM

lib/zaq_web/components/design_system/modal_no_volume.ex
jat10: Not sure we have such a component 
Screenshot 2026-07-30 at 3 49 19 PM

jfayad: we have a modal component and you can put the content you need

lib/zaq_web/components/design_system/modal_skill_resources.ex
jfayad: my point... you can edit the existing component to support defining the volume as a prop/option so it can handle both cases

jfayad (approved):

PR #649, opened 2026-07-30
feat(channels): expose channel provider listing as an agent tool

jat10:


jfayad:
the query resources action covers this

Issue #634, opened 2026-07-27
[refactor][Channels] Isolate from DB


jfayad:
Currently the channels node require an access to the DB, this makes it impossible to run it on a node that isn't connected to an internal networks with DB access.

We should break this dependency and move all DB required access to the engine's node.

For channels' config. we could have an in-memory copy of it so we don't need a DB roundtrip for every message (incoming/outgoing)

So far the identified leaks are:
- Channel's configuration
- Person Identity resolver

- [ ] Identify all leaks and define a clear plan to efficiently move out of Channels node exec. (make sure Repo does NOT start on Channels node)
- [ ] Define phased plan if necessary
- [ ] Implement in independent phases
- [ ] Test by running the channel node on a separate, remote node and confirm it works and scale efficiently

Issue #368, opened 2026-05-07
[AI Agents] Support skills


jfayad:
Jido.AI supports skills.md files spec and this could be leveraged: https://hexdocs.pm/jido_ai/Jido.AI.Skill.html

jfayad:
we should probably also check how to integrate https://agentskill.sh/ so we can list the skills from this directory and have a simple skills discovery --> will do in a separate issue

jfayad:
@jat10 The current implementation of skills does not follow the open agent skills spec and you try to reinvent the wheel instead of using the Jido.Ai.Skill.

The Jido AI implementation contains a few gaps but there are less than your exotic implementation.

Re-align your implementation with Jido AI and eventually open a PR there to fill the gaps: https://github.com/agentjido/jido_ai/issues/323

For a full transcript of the analysis (which I highly invite you to read): https://chatgpt.com/share/6a526ef9-8f98-83eb-82aa-b7db8f5eca61

TL/DR
- You can keep declaration of skills in the DB and build Jido.AI.Skill struct from it to leverage the better Jido.Ai.Skill implementation (which supports progressive disclosure)
- Fix the gap in Jido's global registry before using it
- Declare an action to fully load an activated skill
- Sementically separate the standard allowed_tools key from the custom provided tools one
- custom provided tools and MCPs are ZAQ specific, this should be managed on ZAQ's side
- Leverage the local disk volume for resource attachment and provide a tool for the agent to load requested resources (to keep it progressive)

Read the full transcript to understand the reasoning behind, the gaps and the suggested implementation

jfayad:
**Progress:** Main now has explicit Jido alignment: [spec conversion tests](https://github.com/www-zaq-ai/zaq/blob/a8462624febef7ac8baf1138314d5167800a2bac/test/zaq/agent/skills/spec_test.exs#L33-L74) assert a standard inline-body Spec, stateless resolution without the global registry, allowed_tools, and exclusion of ZAQ-only fields. [Factory coverage](https://github.com/www-zaq-ai/zaq/blob/a8462624febef7ac8baf1138314d5167800a2bac/test/zaq/agent/factory_test.exs#L151-L185) asserts the progressive skill index and load_skill wiring without exposing the body in the initial prompt.

[...]

PR #659, opened 2026-08-01
feat(agent): load skill reference files via unmaterialized records

jat10:


jat10:
# PR Review — #659

| | |
|---|---|
| **Title** | `feat(agent): load skill reference files via unmaterialized records` |
| **PR** | [#659](https://github.com/www-zaq-ai/zaq/pull/659) |
| **Branch** | `fix/skill-resources` → `v0.17` |
| **Linked issues** | [#645](https://github.com/www-zaq-ai/zaq/issues/645) *(closed)* — "[agent]: Support file management for skill references on the BO"; context from [#368](https://github.com/www-zaq-ai/zaq/issues/368) (Support skills) and [#634](https://github.com/www-zaq-ai/zaq/issues/634) (Isolate Channels from DB) |
| **Reviewed at** | 2026-08-03 |
| **Diff** | 40 files, +4040 / −438 |

> The PR body is empty and no commit references an issue. Issue linkage below was
> reconstructed from the branch name and the feature surface. **Recommend adding
> `Closes #645` (or the follow-up issue) to the PR body** so the traceability survives.

---

## Section A — Inline review comments

### `lib/zaq/ingestion/record_materializer.ex`

#### 🔴 High — `delete/2` has no permission check and reports success when the row delete fails

**Lines 113–129.**

```elixir
def delete(%{file_id: file_id}, opts) do
  case fetch_document(file_id) do
    {:ok, doc} ->
      case resolve_full_path(doc.source, opts) do
        {:ok, full_path} -> File.rm(full_path)
        _ -> :ok
      end

      Document.delete(doc)
      :ok
    ...
```

**The problem.** Two defects in one clause:

1. `fetch_document/1` is used, not `fetch_permitted/1`. Every other read path in this
   module goes through `DocumentAccess`; the destructive path does not. `person_id` and
   `team_ids` never even reach here — `DiskBridge.delete_file/2` builds
   `%{file_id: fetch(params, "file_id")}` without `with_caller/2`
   (`lib/zaq/channels/disk_bridge.ex:86-88`), unlike `materialize_request/1` and
   `describe_request/1` which both call it.
2. `Document.delete(doc)`'s return value is discarded and `:ok` is returned
   unconditionally.

**Why it matters.** `delete_file/2` is a callback on the provider-generic
`Zaq.Channels.DataSourceBridge` dispatcher, reachable through the
`:data_source_delete_file` event with `provider: "disk"` and any integer `file_id`. That
call removes the bytes from the ingestion volume *and* the `documents` row for **any**
document in the system, with no caller identity and no check. Today the only caller is
`SkillsLive`, so this is not exploitable from the model — but the module moduledoc asserts
"Every read goes through `Zaq.Ingestion.DocumentAccess`" without saying that writes and
deletes go through nothing, so the next caller (a workflow action, a `delete_document`
tool) inherits arbitrary-document deletion silently.

Defect 2 is reachable today: `SkillsLive.drop_reference/3` treats `:ok` as proof the
document is gone and removes the reference, so a failed `Repo.delete` leaves a row nothing
points at while the operator is told "Resource removed."

**Suggested fix.**

```elixir
def delete(params, opts) do
  with {:ok, doc} <- fetch_permitted(params) do
    case resolve_full_path(doc.source, opts) do
      {:ok, full_path} -> File.rm(full_path)
      _ -> :ok
    end

    case Document.delete(doc) do
      {:ok, _} -> :ok
      {:error, reason} -> {:error, reason}
    end
  else
    {:error, :not_found} -> :ok   # keep idempotence
    error -> error
  end
end
```

and add `with_caller/2` to `DiskBridge.delete_file/2`'s request, matching the other three.
Update the moduledoc's Permissions section to state the rule for writes and deletes
explicitly rather than only for reads.

---

#### 🟡 Medium — `describe/2` drops unresolvable ids silently and returns them out of order

**Lines 61–76.**

```elixir
records =
  ids
  |> permitted_documents(params)
  |> Enum.map(&build_record(&1, nil, opts))
```

**The problem.** `permitted_documents/2` ends in `Repo.all(from d in Document, where: d.id in ^permitted)` (line 149), whose result order is whatever Postgres returns — not the order of `file_ids`. Separately, an id that no longer resolves to a document, or that the caller may not read, is simply absent from `records`; the only trace is `stats.returned < stats.scanned`, which no caller reads.

**Why it matters.** `LoadSkill` renders this list as the skill's `resources` and the BO resource table renders it as the file list. A reference whose document was deleted outside the BO (via the ingestion browser, say) vanishes from both with no warning — the operator sees a shorter table, the model sees a shorter list, and nobody learns the skill has a dangling reference. `ReferenceFiles`' `:on_error` hook only fires when the whole page fails, so it does not cover this. The ordering means the BO table row order can change between renders for no user-visible reason.

**Suggested fix.** Reorder to the requested id order and surface the gap, e.g. return the missing ids in `stats` (`stats: %{scanned:, returned:, missing: ids -- found_ids}`) so `ReferenceFiles` can pass them to `:on_error` and the BO can render a "file no longer available" row instead of nothing.

---

### `lib/zaq/channels/disk_bridge.ex`

#### 🟡 Medium — `node_router` is read out of the caller-supplied params map

**Line 139.**

```elixir
defp dispatch(action, request, params) do
  node_router = fetch(params, "node_router") || NodeRouter
```

**The problem.** `params` is the untrusted half of the `%{provider:, params:}` request that `Zaq.Channels.Api` forwards verbatim from any dispatcher — including agent tools whose params originate with the model. This is the only occurrence of a `"node_router"` params key anywhere in `lib/` (verified by grep); every other module in this PR takes it as an option (`ReferenceFiles.list/2`, `Materializer.materialize/2`, `LoadSkill.load_resources/2` all use `Keyword.get`/`Map.get(context, :node_router, …)`).

**Why it matters.** It is a test seam living in production code, on the request-data path rather than the option path. It is not exploitable from a model today — JSON cannot produce a module atom, and `DownloadDocument` builds its params from a fixed `merge_optional/3` key list — but it makes the dispatch target a function of caller-supplied data, which is exactly the property the rest of the PR is careful to avoid (see the `materializing_event` exclusion rationale in `Zaq.Contracts.Record`'s moduledoc).

**Suggested fix.** `DataSourceBridge` passes `config` as well as `params`; put the seam on `config`, or drop the injection entirely and stub `Zaq.NodeRouter` in the `disk_bridge_test.exs` cases the way `data_source_bridge_test.exs` does.

---

#### 🟢 Low — `get_file/2` trusts that `describe_records` echoed the id it asked for

**Lines 62–68.** `get_file/2` takes `records: [record | _]` from the page without checking `record.id == fetch(params, "file_id")`. It holds today because `describe/2` filters on exactly those ids, but the assertion is implicit across a role boundary. A `record.id` guard costs one line and turns a future contract drift into `:not_found` rather than the wrong file.

---

### `lib/zaq_web/live/bo/ai/skills_live.ex`

#### 🟡 Medium — the resource delete button has no confirmation, unlike every other destructive action on this page

**Template `skills_live.html.heex:366-378`, handler `skills_live.ex:238-256`.**

```heex
<DSButton.button
  variant={:ghost} icon_only type="button"
  aria-label={"Remove #{resource.name}"}
  phx-click="remove_skill_resource"
  phx-value-file_id={resource.id}
```

**The problem.** One click deletes the document row *and* `File.rm`s the file from the ingestion volume — irreversible. The skill delete button 30 lines away in the same template carries `data-confirm=…` (`skills_live.html.heex:440`), and its confirmation text even promises that "its uploaded resources will be removed from the volume" — so the page already establishes that removing a resource file is confirmation-worthy.

**Why it matters.** Accidental single-click data loss with no undo, inconsistent with the established pattern in the same view. Issue #645's requirements section treats deletion as a deliberate act ("deleting a skill that has an associated resource deletes its directory").

**Suggested fix.** Add `data-confirm="Remove this file? It will be deleted from the volume."` to match the skill delete button, or use `BOModal.confirm_dialog/1` per `DESIGN.md`.

---

#### 🟡 Medium — a failed `persist_references/3` is swallowed after the upload toast has already said success

**Lines 540–553.**

```elixir
case save_resources(socket, skill, resources) do
  {:ok, socket} -> socket
  {:error, _reason, socket} -> socket
end
```

**The problem.** The comment above it is accurate about the consequence ("a failure here loses the reference but not the file"), but the operator is never told. `put_resource_upload_toast(uploaded, failed)` runs on the `uploaded` split and reports success, because the *write* succeeded; the reference write that failed afterwards is invisible.

**Why it matters.** The resulting state is an orphan: a file sitting on the volume, tagged `"public"` (so it is retrievable by every agent — see General Comment B2), that no skill references and that `delete_skill_resources/2` will never clean up because cleanup now iterates `Resources.references/1`. The operator's only signal is that the resource table stays empty after a "success" toast, which reads as a rendering bug.

**Suggested fix.** Thread the error into the toast: `{:error, reason, socket} -> put_toast(socket, :error, "Uploaded, but could not attach to the skill: …")`. `save_resources/3` already returns the reason for exactly this purpose — it is just discarded here, while `drop_reference/3` (lines 601–610) does use it.

---

### `lib/zaq/agent/tools/data_source/download_document.ex`

#### 🟡 Medium — no caller identity reaches the `disk` provider, so every read runs anonymous

**Lines 85–102**, in combination with `disk_bridge.ex:92-107` and
`data_source_tool.ex:56-58`.

**The problem.** `DiskBridge.materialize_request/1` calls `with_caller/2`, which reads
`person_id` and `team_ids` out of `params` — but `DownloadDocument.run/2` builds its
request as `%{"file_id" => document_id}` plus `document_mime_type`, `export_mime_type` and
`config_id`, and `DataSourceTool.wrap_request/2` adds nothing. So `person_id` is always
`nil` on this path. `DiskBridge`'s own comment says the opposite is required: "leaving it
off a request would silently downgrade that call to an unprivileged one rather than fail."
That is precisely what happens.

**Why it matters.** Two directions, both wrong:

- **Under-permissioned:** a document the asking person legitimately may read, via a
  `Permission` row, is unreachable through the tool. Only `"public"`-tagged documents
  resolve, which is why `SkillsLive` has to tag every skill reference `"public"` at write
  time (`@resource_tags`). The tag is a workaround for the missing identity, and it is a
  global one.
- **Over-permissioned:** `document_id` is a model-chosen string mapped straight to a
  sequential `documents.id`. Any `"public"` document in the install is fetchable by
  guessing an integer, which bypasses whatever scoping retrieval applies (source filters,
  tags, folder settings). A prompt-injected agent can walk `1..N`.

**Suggested fix.** Propagate the caller from the tool context the way the rest of the agent
tool surface does — `Map.get(context, :person_id)` / `:team_ids` into the request map — and
add `person_id` / `team_ids` to `DataSourceTool.wrap_request/2` so every datasource tool
carries it rather than each one remembering. Once identity flows,
`SkillsLive.@resource_tags` can become a skill-scoped permission row instead of the global
`"public"` tag.

---

### `lib/zaq/agent/skill.ex`

#### 🟢 Low — the comment says `"skills"` and `"assets"` are reserved; the code rejects them

**Lines 141–145 and 157–161.**

```elixir
# `resources` is a namespace map. Only `"references"` is populated today; `"skills"` and
# `"assets"` are reserved so they can be added without another migration.
...
@resource_kinds ~w(references)

defp validate_resource_map(changeset, resources) do
  case Map.keys(resources) -- @resource_kinds do
    [] -> ...
    unknown -> add_error(changeset, :resources, "has unknown keys: #{Enum.join(unknown, ", ")}")
```

Writing `%{"assets" => []}` is a validation error, not a reserved-but-empty namespace. The
migration comment repeats the claim. Either add them to `@resource_kinds` (and have
`validate_references/2` stop clobbering the other keys — line 165 does
`put_change(:resources, %{"references" => …})`, which drops anything else present), or
reword both comments to say the *column shape* is what leaves room, not the validator.

---

#### 🟢 Low — reference de-duplication ignores the provider half of the address

**Line 175**, and `resources.ex:70-79, 86-91`.

```elixir
defp dedupe_references(references), do: Enum.uniq_by(references, &Map.get(&1, "file_id"))
```

The module's own docs define a reference as `{file_id, provider}` — "`id` and `provider`
together are the address `download_document` takes" (`load_skill.ex:37`). De-duping on
`file_id` alone means `{"42", "disk"}` and `{"42", "google_drive"}` collapse to one, and
`Resources.remove_reference/2` removes both. Document ids are provider-scoped, so the
collision is realistic once a second provider is wired up. `Enum.uniq_by(&{&1["file_id"],
&1["provider"]})` in all three places matches the stated model.

---

### `docs/exec-plans/tech-debt-tracker.md`

#### 🟢 Low — duplicated heading and a dangling plan reference

Two consecutive `### Should Do` headings under the new `## Ingestion` section; the second
should be a plain list item under the first.

More importantly, the first entry points at
`docs/exec-plans/active/skill-reference-materialization.md`. That file exists locally but
is **untracked and not part of this PR** (`git ls-files docs/exec-plans/active/` is empty).
Every other reader follows a dead link. Either commit the plan with the PR or drop the
reference.

---

## Section B — General PR conversation comments

### B1 🟡 Medium — the PR body is empty and nothing links it to an issue

No `Closes #…`, no issue number in any of the five commit subjects, no branch-name
convention that resolves to one. #645 is already **closed**, so this PR reads as
unattributed work. For a 4000-line change that alters the permission surface of ingestion
(new `:ingestion` role actions, a new datasource provider, a schema migration), the
missing "why now" is a real maintenance cost.

**Recommendation.** Add a body stating: the issue(s) it advances, that
`resource_root` → `resources` is a breaking schema change with no backfill and why that is
safe (the feature has not shipped), and the two behaviour changes called out in B2 and B3.

---

### B2 🟠 Medium-High — skill reference files are made globally public, and this deserves an explicit decision

`SkillsLive.@resource_tags = ["public"]` (line 65) tags every uploaded reference file with
the same `"public"` tag ingestion uses for install-wide readability. The moduledoc is
admirably honest about the consequence: *"It also makes the file visible in ingestion
browse and ordinary retrieval, not only through `load_skill`."*

Concretely: a file uploaded as a reference for skill A is readable by **every** agent
(through `download_document` with a guessed id), surfaces in **every** RAG retrieval, and
is browsable by any BO user with ingestion access. Nothing scopes it to the skill.

Issue #645 says nothing about the files being install-wide readable; the screenshots in the
comment thread present them as skill-scoped attachments, which is how an operator will read
the UI. An operator uploading a pricing-FAQ PDF to one skill will not expect it to appear
in unrelated retrieval results.

**Evidence.** `skills_live.ex:60-65`; `record_materializer.ex:16-24`;
`document_access.ex:45-52` (`via_public` matches on the `"public"` tag with no scope).

**Recommendation.** This is a product decision, not a code nit — but it should be a
recorded one. Either (a) note it in the PR body and open a follow-up issue to replace the
tag with a skill-scoped `Permission` row once caller identity flows (see the inline finding
on `download_document`), or (b) exclude `.agents/skills/**` sources from retrieval so the
tag only grants direct-id reads. Silently shipping "skill attachment" as "install-wide
public document" is the outcome to avoid.

---

### B3 🟢 Low — `resolve_data_source_config/2` now succeeds for every unconfigured provider

`data_source_bridge.ex:489-512` changes the no-`channel_configs`-row case from
`{:error, …}` to `{:ok, %{provider: to_string(provider)}}` for **all** providers, not just
`disk`. The compensating guard is one new fallback clause,
`JidoConnectBridge.runtime_ctx/1` (line 1392).

I verified the surface: every `JidoConnectBridge` datasource callback
(`list_files`, `create_file`, `get_file`, `update_file`, `delete_file`, `search_files`,
`download_document`, `list_permissions`, `export_options`, `channel_stats`, all seven
`sheet_*`) funnels through `invoke_intent/3` → `runtime_ctx/1`, so the guard is complete
**today**. Two notes rather than defects:

- Nothing enforces that invariant. A future callback that reads `config.id` before
  reaching `runtime_ctx/1` (there are 18 such sites in the file) turns a clean
  `{:channel_not_configured, provider}` into a `FunctionClauseError`. Worth a line in the
  `DataSourceBridge` moduledoc stating the contract: *a bridge must tolerate a config with
  no `:id`.*
- `export_options/2` swallows errors into `default_export_options_response()`
  (`jido_connect_bridge.ex:283-300`), so an unconfigured provider now answers with default
  export options instead of an error. Almost certainly harmless; flagging because it is a
  behaviour change nobody asked for.

---

### B4 ⭐ Standout — the `materializing_event` exclusion is genuinely well designed

Worth calling out because it is the kind of thing that is usually got wrong. `Record`
excludes `materializing_event` from *one* `@public_fields` list shared by both the
`Jason.Encoder` derivation and `to_map/1` (`record.ex:16-46, 107`), with the reasoning
written down: a dispatchable event inside a serialized payload would let whoever holds the
record choose which event fires, on which node, with which params. The consequence — a
record rebuilt from an LLM tool result can never be materialized — is stated in three
places and is load-bearing. `DownloadDocument`'s `@output_schema` comment
(`download_document.ex:41-45`) explains why `Zoi.map` rather than `Zoi.struct` and ties it
back to the same list. This is the right shape.

The `carry_over/2` vs `merge_payload/2` pair (`materializer.ex:164-183`) is also good:
same field list, deliberately opposite precedence, with a comment saying so rather than
one function with a flag.

---

### B5 🟢 Low — test coverage is strong; two gaps

~1900 lines of new tests against ~1200 lines of new production code, including a real
multi-hop e2e (`skill_reference_e2e_test.exs`) rather than a stubbed seam — consistent with
`docs/testing-approach.md` and with the "never stub the hop under test" rule. All 364
tests across the changed suites pass locally (`0 failures`).

Two paths I could not find a test for:

- `RecordMaterializer.delete/2` when `Document.delete/1` fails — which is the second half
  of the High finding above, and is untestable-as-specified because the function cannot
  report it.
- `SkillsLive.persist_references/3` when `save_resources/3` returns `{:error, …}` — the
  orphan-file path in the Medium finding. `skills_live_test.exs` covers the success and the
  `drop_reference` error path but not this one.

---

## Section C — Final recap

### 1. Overall verdict

**Request changes** — for the `RecordMaterializer.delete/2` finding alone. Everything else
is addressable in review; that one deletes data with no permission check and reports
success when it fails.

### 2. Summary

This is high-quality work. The architectural core — an unmaterialized `Record` carrying a
`materializing_event`, dispatched by a dedicated `Materializer` that resolves both
one-hop and redirecting providers — is a genuinely good abstraction, and it is documented
at the level of *why* rather than *what*, in line with `docs/conventions.md`. Moving skill
references from a path (`resource_root`) to a `{file_id, provider}` pair is the right call
and correctly kills the rename-orphans-a-file class of bug. The `disk`-as-datasource
framing lets the BO write, tag and register a file in one dispatch instead of three, and
directly advances #634 by keeping the volume knowledge on the ingestion node. Tests are
thorough and use real seams.

**Does it resolve the linked issue?** #645's functional requirements are met and then some
— upload, staging before the skill exists, volume switching, deletion on skill delete, and
now per-file removal, which #645 did not ask for. The reviewer comments on #645 (file-type
limits `md/pdf/png/json`, placement between Allowed tools and Instructions) are satisfied
by `@allowed_extensions` and the existing template.

**Key risks.** (1) The unguarded delete path. (2) No caller identity reaches the `disk`
provider, which forces the global `"public"` tag and makes every skill reference file
install-wide readable and retrievable — a product consequence that is documented in code
but not surfaced anywhere a decision-maker will see it. (3) Two silent-failure paths in the
BO that leave orphaned public files after a "success" toast.

**Positives worth keeping.** The `materializing_event` exclusion design (B4); the
`@public_fields` single-list discipline; the honest tech-debt entries filed rather than
worked around silently; the `FilePreviewData` text-sniffing fix, which is a correct
narrow guard with the general fix filed as debt.

### 3. Findings summary

| Severity | Count |
|---|---|
| 🔴 High | 1 |
| 🟠 Medium-High | 1 |
| 🟡 Medium | 5 |
| 🟢 Low | 6 |
| ⭐ Positive | 1 |
| **Total** | **14** |

- **Inline comments:** 9 (1 high, 4 medium, 4 low)
- **General comments:** 5 (1 medium-high, 1 medium, 2 low, 1 positive)

### 4. Coverage summary

**Issues inspected:** #645 (full description + both comments, recursively — the
`jfayad` → `jat10` exchange on file types and UI placement was checked against the
implementation). **Related issues inspected:** #368 (Support skills — parent epic),
#634 (Isolate Channels from DB — the architectural motivation for `disk`-as-datasource).
Searched the full issue list for `materializ`, `disk bridge volume`, `skill resource`,
`skill reference files`; no other related issue exists.

**Files reviewed in full:** `contracts/record.ex`, `contracts/record/materializer.ex`,
`ingestion/record_materializer.ex`, `channels/disk_bridge.ex`,
`agent/skill/reference_files.ex`, `agent/skill/resources.ex`, `utils/map.ex`,
`agent/tools/skills/load_skill.ex`, `agent/tools/data_source/download_document.ex`.
**Reviewed as diff:** `agent/skill.ex`, `agent/skills.ex`, `channels/data_source_bridge.ex`,
`channels/jido_connect_bridge.ex`, `ingestion/api.ex`, `ingestion/record_source.ex`,
`bo/ai/skills_live.{ex,html.heex}`, `bo/ai/file_preview_data.ex`,
`agent/tools/resources/registry.ex`, `config/config.exs`, the migration, and
`docs/exec-plans/tech-debt-tracker.md`.
**Untouched code inspected to verify findings:** `ingestion/document_access.ex`,
`ingestion/file_explorer.ex` (`resolve_path/2` traversal containment — the `Resources`
moduledoc's claim that "ingestion rejects traversal independently" **holds**),
`channels/api.ex` (datasource event handlers), `agent/tools/data_source_tool.ex`,
`agent/tools/data_source/` (full tool inventory, to establish the model-reachable surface
for `provider: "disk"`).

**Risk areas checked:** permission enforcement on every new `:ingestion` role action
(`materialize_record`, `describe_records`, `persist_record`, `delete_record`); path
traversal on the upload path; capability leakage through JSON/`to_map` serialization;
the model-reachable `disk` surface via all six `data_source` tools; the
`resolve_data_source_config/2` behaviour change against every `JidoConnectBridge`
callback; N+1 / redundant dispatch (none found — `ReferenceFiles.list/2` correctly groups
to one dispatch per provider); duplicate code introduced (none — commit `9a9f8b8b` in fact
*removes* duplication by extracting `ReferenceFiles`); duplicate LiveView markup (none);
`NodeRouter` discipline (all cross-service hops use `NodeRouter.dispatch/1` with
`%Zaq.Event{}`); moduledoc coverage on new modules (complete, and well above the project
baseline).

**Verification run:** `mix test` over the 12 changed/new suites — **364 tests, 3
properties, 0 failures**.

### 5. Token usage

Runtime does not expose exact per-turn token metrics; the following are estimates derived
from the volume of file content and command output pulled into context.

| Metric | Estimate |
|---|---|
| Input tokens (cumulative across ~18 tool-calling turns) | ~270,000 |
| Output tokens | ~11,000 |
| **Total** | **~281,000** |

Largest contributors: the four fully-read new modules (~10k), the five diff hunks (~18k),
and the cumulative re-send of conversation context across turns (the dominant term).

PR #696, opened 2026-08-22
feat(engine): derive workflow trigger input contract with validation tool

### LINKED ISSUE #697: [workflow]: Provide a tool for agents to validate the input to any workflow
Some workflow work based on events (Zaq.Events) that get triggered from different sources it could be from workflows or any function in the code. 

# Problem:
- If we want to let agents to build this input we need a tool that validates the input that get build, that whenever it get triggered we need to a guarantee that the workflow will not fail because of missing input.

# Solutions:
We should validate to dynamically load the inputs of a workflow and make sure the workflow will run before we trigger it:
SUM of all inputs - SUM of inputs that are fed by the output of a previous step (except start)

jat10:
In this PR:
- Adds `Zaq.Engine.Workflows.InputContract`, which derives what a workflow's trigger payload must contain by subtracting the inputs upstream steps feed from every input its nodes need, reading edge mappings, {{start.*}} params, edge conditions, and action schemas.
- Exposes it to agents as the `workflow.validate_workflow_input` tool, which checks a candidate payload before dispatch and returns the gaps plus a fillable nested shape to correct them.

Fix #697


test/zaq/engine/workflows/real_workflow_input_contract_test.exs
```diff
@@ -0,0 +1,355 @@
...
+
+  @fixtures_dir Path.join(__DIR__, "../../../support/fixtures/workflows")
+
+  # ── Generate Company Context — 28036805 ─────────────────────────────────────
+
+  # `company context content` is the branch selector: empty routes through the
+  # extract/summarise cascade, non-empty dispatches `craft_email_direct` straight
+  # away. Both branches read it, so the payload must carry it either way.
+  @generate_company_context_input %{
+    "company website" => "https://www.oilicraft.com/",
```
jfayad: remove all such references

jfayad:
/zaq You are a principal-level software reviewer acting as a PR review agent.

Your goal is to produce a high-signal review that is accurate, contextual, and actionable.

## Primary review objectives
Review the pull request with a balanced approach focusing on:

1. Code quality and best practices
- naming clarity
- readability
- cohesion and separation of concerns
- duplication across modules
- adherence to established project patterns
- consistency with surrounding code style

2. Performance
- inefficient algorithms
- avoidable allocations
- unnecessary renders or recomputation
- redundant DB/API calls
- N+1 access patterns
- excessive serialization/deserialization
- memory retention risks
- unnecessary processing in hot paths

3. Logic and correctness
- broken flows
- incorrect assumptions
- edge cases
- race conditions
- missing guards/validation
- off-by-one and boundary bugs
- incorrect error handling
- regressions relative to the issue being fixed

4. Maintainability
- missing or weak module/function documentation where it materially hurts maintainability
- code that should be extracted into reusable helpers/components
- duplicate LiveView UI code that should become reusable components
- hidden coupling
- weak testability

## Required issue-context review
Before reviewing the code, you must gather and use the issue context that the PR is intended to fix.

You must:
- identify the issue(s) referenced by the PR body, branch name, title, commits, linked references, or closing keywords
- read the full issue description
- read issue comments
- recursively inspect directly related issues when they are clearly referenced and relevant
- stop recursion when additional issues become only loosely related or duplicate prior context

Use the issue context to check whether:
- the PR actually solves the reported problem
- important acceptance criteria are missing
- edge cases raised in issue comments were ignored
- the implementation introduces behavior inconsistent with the issue intent

## Review policy
- Prefer substantive findings over stylistic nitpicks
- Skip purely cosmetic comments unless they materially affect readability or maintenance
- Be conservative: do not invent problems without evidence from the code, diff, or issue context
- Do not flag project-specific style choices if they are already consistently used in the surrounding codebase
- Call out standout positive changes when they are meaningful

## Output requirements
You must produce three outputs:

### A. Inline review comments
When a finding maps to a line or hunk that is part of the PR diff, produce an inline review comment suitable for posting on that file/line.

Each inline comment must include:
- severity: high | medium | low
- file path
- diff line reference if known
- concise title
- the problem
- why it matters
- a concrete suggested fix

Keep inline comments tight and directly tied to the changed code.

### B. General PR conversation comments
When a valid finding cannot be attached to a changed hunk, produce a general PR conversation comment.

Use this for:
- architectural concerns
- missing tests not tied to one diff line
- issue-scope mismatch
- related code outside the diff
- cross-module duplication
- undocumented behavior changes
- acceptance-criteria gaps

Each general comment must include:
- severity
- title
- the problem
- supporting evidence
- a concrete recommendation

### C. Final recap
Produce a recap with:
1. overall verdict
   - approve
   - comment
   - request changes

2. summary
   - short assessment of code quality
   - whether the PR appears to resolve the linked issue(s)
   - key risks
   - noteworthy positives

3. findings summary
   - count by severity
   - count of inline comments
   - count of general comments

4. coverage summary
   - issues inspected
   - related issues inspected
   - files reviewed
   - major risk areas checked

5. token usage
   - estimated input tokens
   - estimated output tokens
   - estimated total tokens
   - if available, actual token metrics from the runtime/tooling

## Auditing checklist
You must explicitly check, within the scope of the PR only:
- duplicate code introduced or expanded by the PR that should be centralized into reusable functions
- duplicate LiveView code introduced or expanded by the PR that should become reusable components
- mismatching coding style in changed code relative to the local codebase
- performance problems introduced, preserved in the changed path, or made more likely by the PR
- lacking module/function documentation in touched code where documentation would materially improve maintainability
- whether the PR truly addresses the linked issue and issue comments

## Evidence standard
Only report a finding when at least one of these is true:
- you can point to a concrete changed line
- you can point to concrete surrounding code in the repo
- you can point to a requirement or scenario from the linked issue(s)

## Severity guidance
- High: likely bug, broken flow, security/reliability problem, major performance issue, or PR does not actually solve the issue
- Medium: meaningful maintainability/performance/correctness risk
- Low: smaller but worthwhile issue with real practical value

## Review process
Follow this order:
1. read PR metadata and diff
2. determine linked issue(s)
3. read issue descriptions/comments and relevant related issues
4. inspect changed files and nearby code
5. inspect related untouched code where needed
6. generate inline comments
7. generate general comments
8. generate final recap
9. include token statistics

## Constraints
- Do not modify code
- Do not post duplicate comments
- Do not restate the same issue inline and in the general thread unless there is a strong reason
- Keep comments crisp, specific, and actionable
- Avoid praise-only noise; praise only when it highlights something meaningfully well done

jat10:
/zaq evaluate the changes pushed in the PR AFTER the review and confirm all comments shared have been addressed correctly

jfayad:
/zaq evaluate the changes pushed in the PR AFTER the review and confirm all comments shared have been addressed correctly

jat10:
For the last Zaq's Review:

1. This a design decision since the purpose of this action is to validate the input and not the values. So a key with nil value is valid. 
2. Not a bug `{{...}}` substitution and its "" collapse are the documented contract (Concat and friends rely on empty string, not nil), and the key-collision case is the accepted trade-off already noted at step_runner.ex:251-253, which this PR narrows rather than introduces.
3. Since this is not yet into production there is no need for migration. We have only one workflow that requires a change which is `send_leads_email.json` it's a one line `"input": "{{build_history.metadata}}",`

jfayad:
> For the last Zaq's Review:
> 
> 1. This is made like this in design since the purpose of this action is to validate the input and not the values. So a key with nil value is a valid.

When was this only about the input's shape and not about the input's value ?

The short brief for this action is literraly
> We should validate to dynamically load the inputs of a workflow and make sure the workflow will run before we trigger it

the finding is stating that you are letting a situation that is known to fail to pass this check

> 2. Not a bug `{{...}}` substitution and its "" collapse are the documented contract (Concat and friends rely on empty string, not nil), and the key-collision case is the accepted trade-off already noted at step_runner.ex:251-253, which this PR narrows rather than introduces.

My previous observations were that similar data provided by a later step would not override the previous data and all is still accesible by scoping access with the step's node key (i.e NodeA.name and NodeB.name are both accessible), the finding seem to state that this isn't the case ?

> 3. Since this is not yet into production there is no need for migration. We have only one workflow that requires a change which is `send_leads_email.json` so this can and it's a one line `"input": "{{build_history.metadata}}",`

You are talking about yourself, but other users have enabled workflows, this feature still being in beta a full migration is not necessary, but proper communications is necessary!

jat10:
Reporting the breaking change:

**1. A Condition's `input` reference must be `{{...}}`.** A bare dotted string (`"input": "build_history.metadata"`) used to resolve against the cascade; it is now literal data, so every condition misses and the node never fires. Write `"{{build_history.metadata}}"`.

**2. A Condition's `conditions[].key` no longer reaches the cascade.** A key is now a selector into `input` only, so `"start.language"` and `"store_context.record.id"` stop resolving — the condition falls to its `default` and quietly takes the other branch. Bring the value into `input` first, then address it locally.

**3. `RunAgent` no longer spreads nested params one level.** `{{email}}` against a `row` param now resolves to `""`; write the path, `{{row.email}}`. Applies to `input` and to each context turn's `content`.

jfayad:
I cannot confirm the branch fully implements the requested features.

## 1. **High: No function exposes the starting input as a Zoi schema.** 

`InputContract` exposes lists, textual types, a nil-valued shape, and a verdict map at `lib/zaq/engine/workflows/input_contract.ex:69-215`. `Workflows.input_contract/2` only returns the verdict at `lib/zaq/engine/workflows.ex:397-415`.

## 2. **High: The inferred contract can omit legitimate inputs.**
   - Optional schema fields are only included when explicitly wired/referenced; `schema_needs/4` enumerates required fields only at `input_contract.ex:568-620`.
   - Composed child workflows are not expanded before derivation, unlike runtime expansion at `workflows.ex:643-653`.

### Bare sources on `start` mappings can be classified as unsatisfiable at `input_contract.ex:623-650`.

For this edge:

```elixir
%{
  "from" => "start",
  "to" => "notify",
  "mapping" => %{"message" => "payload_message"}
}
```
Runtime behavior is valid:

- `EdgeStep` receives the initial payload as its root fact.
- `FactLookup.fetch(fact, "payload_message")` reads that top-level payload key (`fact_lookup.ex:62-66`).
- The value becomes `message`.

Contract behavior differs:

- `edge_needs/3` discards the edge’s `"from"` before classifying the source (`input_contract.ex:544-548`).
- `need/4` recognizes only explicit `"start.*"` sources as trigger inputs (`input_contract.ex:623-630`).
- Bare `"payload_message"` is passed through `root_kind/2` and will normally become `:unsatisfiable` (`input_contract.ex:642-650`).
- Unsatisfiable needs are not exposed as required payload paths.

So the contract can omit a payload field that runtime successfully reads. Writing `"start.payload_message"` avoids the bug, but runtime supports both forms. The correct classification needs the edge origin: every mapping source on an edge whose `from == "start"` should be treated as trigger input.

### Mapping-consumed predecessor fields are still considered available by `input_contract.ex:535-542`, although runtime removes them at `steps/edge_step.ex:181-191`.

the predecessor value remains accessible through `node.field`.

`StepRunner.inject_cascade/3` preserves it under the node namespace (`step_runner.ex:229-231`). `EdgeStep` only removes the unqualified root key (`edge_step.ex:181-191`).

For example, after:

```elixir
A returns: %{email: "a@example.com"}
A -> B mapping:
%{"recipient" => "email"}
```
The downstream fact is conceptually:

```elixir
%{
  recipient: "a@example.com",
  __cascade__: %{a: %{email: "a@example.com"}}
}
```
Therefore `a.email` still works in any successor on that path.

The narrower problem is that accessibility does not automatically populate B’s action parameter named `email`. If B requires both `recipient` and `email`, the contract sees `email` in A’s output schema through `upstream_emits/2` and considers it supplied (`input_contract.ex:535-542,590-598`). At runtime, however, root `email` was consumed. B receives no `email` unless explicitly wired with:

```elixir
%{"email" => "{{a.email}}"}
```
or:

```elixir
%{"email" => "a.email"}
```

So the finding is valid only for automatic unqualified schema-field propagation. Node-qualified access itself is preserved.

## 3. **High: The agent-callable endpoint has no authorization boundary.** 
`Engine.Api` ignores context at `lib/zaq/engine/api.ex:579-586`, while the tool forwards neither actor nor permission context at `validate_workflow_input.ex:135-138`.

## 4. **High: A valid verdict does not guarantee runtime success.** 
Defaulted Zoi fields are considered optional at `action.ex:344-351`, but `StepRunner` invokes actions without applying defaults at `step_runner.ex:195-205`. For example, `EncodeBase64.run/2` requires defaulted keys in its function head at `encode_base64.ex:53-54`.

## 5. **Medium: Errors are not fully standardized.** 
Empty input returns a prose `{:error, "input is required"}` rather than validation errors, and the declared `errors` schema is only an unconstrained map array at `validate_workflow_input.ex:65-87`.

## 6. **Medium: Legacy residue remains.** 
The old per-action resolvers are removed, but `RunAgent` still documents and advertises direct substitution at `run_agent.ex:3-69` and `registry.ex:334-338`, while direct calls no longer substitute. Workflow docs also claim `StepRunner` performs runtime schema validation at `docs/services/workflows.md:190-199`, but that validation was removed.

**Tests**

All 13 current PR checks pass. However, coverage is missing for Zoi-schema output, composition, bare start mappings, consumed mapping sources, authorization, and exact placeholder-value collisions. Some tests explicitly codify omitted optional fields and default-related behavior rather than detecting these gaps. I did not rerun tests locally because this review was performed under read-only plan mode.

Issue #761, opened 2026-09-17
[workflow]: Placeholder substitution loses provenance


jat10:
### Exact placeholder-value collisions can corrupt runtime data.
Runtime values equal to the original authored placeholder are incorrectly treated as authored and substituted.
Reference: `step_runner.ex:274-278`.

Replacing StepRunner mod.run with Jido.Exec.run would address the schema/defaults bug, but not the placeholder-collision bug

`lib/zaq/engine/workflows/step_runner.ex:196-207` — the only two call sites:

```elixir
defp call_action(mod, action_params, context, nil), do: mod.run(action_params, context)
defp call_action(mod, action_params, context, timeout_ms) do
  task = Task.async(fn -> mod.run(action_params, context) end)
  ...
```

Jido.Exec.run would give us the ability to enforce providing the action the required input because it does the validation while action.run doesn't do the checks. 

Here an example

```elixir
iex(1)> Jido.Exec.run(ValidateShipment, %{label: "shipment-1", quantity: 10,
                                           weight_kg: 2.5, notify_email: "omple.com"})
{:error,
 %Jido.Action.Error.InvalidInputError{
   message: "is required, at reference_code\ninvalid email format, at notify_email",
   field: nil,
   value: nil,
   details: %{
     module: Zaq.Agent.Tools.Workflow.ValidateShipment,
     context: "Action",
     errors: [
       %{code: :required, message: "is required", path: [:reference_code]},
       %{
         code: :invalid_format,
         message: "invalid email format",
         path: [:notify_email]
       }
     ]
   }
 }}
```

jfayad:
**Reproducer clarification:** Could you attach the minimal workflow definition, runtime input containing the exact authored-placeholder string, and expected literal output for this collision? On merged main `a8462624`, the [cited StepRunner region now persists action results](https://github.com/www-zaq-ai/zaq/blob/a8462624febef7ac8baf1138314d5167800a2bac/lib/zaq/engine/workflows/step_runner.ex#L245-L278) rather than performing the referenced substitution, so the old line reference does not identify the current failing path. Please also identify the commit or resolving layer where the collision was observed. This is not a claim that the bug is fixed; it will keep the provenance regression distinct from schema/default handling in #706. No tests were run.

PR #765, opened 2026-09-18
fix(workflow): action execution through Jido.Exec

### LINKED ISSUE #706: [workflow]: Validate step input and feed defaults
### Defaults are not applied before action execution.
`StepRunner` invokes `mod.run/2` directly, so actions such as `EncodeBase64` can crash when optional defaulted fields are omitted.
References: `step_runner.ex:195-205`, `encode_base64.ex:53-54`.

### Exact placeholder-value collisions can corrupt runtime data.
Runtime values equal to the original authored placeholder are incorrectly treated as authored and substituted.
Reference: `step_runner.ex:274-278`.

Replacing StepRunner mod.run with Jido.Exec.run would address the schema/defaults bug, but not the placeholder-collision bug

`lib/zaq/engine/workflows/step_runner.ex:196-207` — the only two call sites:

```elixir
defp call_action(mod, action_params, context, nil), do: mod.run(action_params, context)
defp call_action(mod, action_params, context, timeout_ms) do
  task = Task.async(fn -> mod.run(action_params, context) end)
  ...
```

Jido.Exec.run would give us the ability to enforce providing the action the required input because it does the validation while action.run doesn't do the checks. 

Here an example

```elixir
iex(1)> Jido.Exec.run(ValidateShipment, %{label: "shipment-1", quantity: 10,
                                           weight_kg: 2.5, notify_email: "omple.com"})
{:error,
 %Jido.Action.Error.InvalidInputError{
   message: "is required, at reference_code\ninvalid email format, at notify_email",
   field: nil,
   value: nil,
   details: %{
     module: Zaq.Agent.Tools.Workflow.ValidateShipment,
     context: "Action",
     errors: [
       %{code: :required, message: "is required", path: [:reference_code]},
       %{
         code: :invalid_format,
         message: "invalid email format",
         path: [:notify_email]
       }
     ]
   }
 }}
```

jat10:
## Summary

* Replace direct `mod.run/2` calls with `Jido.Exec` for wrapped workflow actions.
* Apply Jido schema validation, defaults, hooks, output validation, retries, and timeouts consistently.
* Normalize JSONB/string-key params with `Jido.Action.Tool.convert_params_using_schema/2`.
* Change HITL from error-based waiting to typed `PendingApproval` success metadata.
* Intercept pending results before Runic applies downstream DAG effects.
* Handle suspension correctly inside map/fork execution.
* Align `Condition` schema with its actual supported inputs.
* Remove ZAQ’s duplicate `Task.async`, timeout, and retry machinery.
* Keep ZAQ responsible for workflow state/persistence; Jido owns action execution.
* Add regression coverage for the real workflow -> `Jido.Exec` execution seam.

fix #706


lib/zaq/engine/workflows/workflow_run_agent.ex
```diff
@@ -165,16 +162,85 @@ defmodule Zaq.Engine.Workflows.WorkflowRunAgent do
...
+        )
+
+        reraise exception, __STACKTRACE__
     catch
       :throw, :pause_requested ->
         Logger.info("[workflow] run paused", run_id: run.id)
-        RunWatcher.done(watcher)
         {:ok, Workflows.get_run!(run.id)}
+    after
+      RunWatcher.done(watcher)
```
jfayad: **[medium] Keep orphan recovery armed when the driver exits unexpectedly**

`after` runs for an internal `exit/1`, but `rescue` handles only exceptions and the catch clause only handles `:pause_requested`. Consequently an infrastructure exit sends `:done` to the watcher without persisting interruption. The watcher terminates normally instead of handling the driver's DOWN signal. Its own module documentation explicitly warns against this blanket cleanup.

Reproduced with an internal exit from a native Runic step: after both the driver and watcher terminated, the run still had **`status: "running"`**. A DBConnection checkout exit at a driver checkpoint is another applicable source. Existing recovery tests inject external `Process.exit/2`, which does not test this stack-unwinding path.

Signal `done` only after a durable normal/recovered outcome, or explicitly handle all failure kinds before disarming. Add an internal-exit regression that monitors both processes and verifies persisted run/step recovery.

lib/zaq/agent/tools/workflow/condition.ex
```diff
@@ -60,19 +60,19 @@ defmodule Zaq.Agent.Tools.Workflow.Condition do
...
-            "must keep its own `input` to fire. When absent — e.g. a Condition that is the " <>
-            "first node off a trigger — `run/2` falls back to the incoming fact at root; " <>
-            "`start.<field>` dotted keys reach the trigger payload."
+            "must keep its own `input` to fire. Input is required, including trigger-first " <>
+            "conditions: use an explicit map or a reference such as `start`. " <>
+            "Unresolved or non-map references are validation errors."
       ],
       conditions: [
-        type: {:list, :map},
+        type: {:list, {:map, :any, :any}},
```
jfayad: **[medium] Preserve an object schema for Condition's agent-tool inputs**

The pinned Jido JSON-schema converter recognizes `:map` and `{:map, _}`, but not the three-element `{:map, :any, :any}` form (`deps/jido_action/lib/jido_action/schema.ex:389–403`). This change makes the registered `workflow.condition` tool advertise `conditions` as an array of **strings**, although runtime validation requires maps.

Confirmed with `Jido.Action.Tool.build_parameters_schema(Condition.schema())`: `properties.conditions.items.type` is `"string"`, not `"object"`. An agent following the published schema produces arguments the action rejects.

Use a schema representation supported by both runtime validation and JSON-schema generation, or fix the shared converter through the dependency's supported mechanism. Add a generated-tool-schema assertion and a tool-execution contract test; direct Condition validation tests miss this consumer.

lib/zaq/engine/workflows/workflow_run_agent.ex
```diff
@@ -165,16 +162,85 @@ defmodule Zaq.Engine.Workflows.WorkflowRunAgent do
...
         end
 
-      RunWatcher.done(watcher)
       outcome
+    rescue
+      exception ->
+        Workflows.interrupt_run(run,
+          reason: "driver_crash",
+          message: Exception.message(exception)
+        )
```
jfayad, line 179: **[medium] Do not interrupt an outcome that finalization already committed**

This rescue also covers `finalize/2` and its post-update notifications, but passes the earlier `run` struct whose status is still `running`. `interrupt_run/2` checks that struct, not the current database status. Reproduced by raising only on the `run.completed` dispatch: the step is durably completed, yet the persisted run is overwritten to **`interrupted`**. A notification failure after the waiting transition can likewise make a pending approval inaccessible.

Keep lifecycle notification failures from rewriting execution state, and condition interruption on the current persisted active state rather than a stale struct (with a race-safe transition). Add regressions for dispatch failure after completed/waiting commits. This also violates the module/service contract that lifecycle-event dispatch failures must not affect run state.

lib/zaq/engine/workflows/map_node_builder.ex
```diff
@@ -164,12 +179,14 @@ defmodule Zaq.Engine.Workflows.MapNodeBuilder do
   defp run_fork(fact, specs) do
     specs
     |> Enum.reduce_while({:ok, fact}, fn spec, {:ok, prev} ->
-      case StepRunner.run(Map.merge(prev, spec), %{}) do
+      case Jido.Exec.run(StepRunner, Map.merge(prev, spec), %{}, ExecutionPolicy.outer_options()) do
+        {:ok, %{}, workflow_control: %PendingApproval{} = control} -> {:halt, {:pending, control}}
         {:ok, result} -> {:cont, {:ok, result}}
         {:error, _} -> {:halt, {:error, map_index_of(prev)}}
```
jfayad: **[high] Do not turn an unpersisted wrapper failure into successful item loss**

The outer `Jido.Exec` now catches failures that occur before StepRunner records a failed cursor, but this branch still assumes every error has a durable failed row. Reproduced with a one-item map using `fail_workflow`: create the HITL approval, pause before the waiting write, then call public `resume_run/1`. `pause_run` leaves the cursor `paused`; `recover_approval_cursor` rejects it, and the resulting MatchError becomes this error sentinel. FanIn drops the item and MapCollect finds no failed row. The run finishes **`completed`, `count: 0`, `errors: []`**, while its approval remains pending.

Recover legitimate paused pending-approval cursors, and propagate or durably record outer-wrapper/infrastructure errors instead of silently treating them as already-accounted-for fork failures. Add a public pause/resume regression asserting that the item remains suspended until decided and cannot disappear even under `fail_workflow`.

lib/zaq/engine/workflows/step_runner.ex
```diff
@@ -243,6 +262,7 @@ defmodule Zaq.Engine.Workflows.StepRunner do
     prev_cascade = Map.get(params, :__cascade__, Map.get(params, "__cascade__", %{}))
 
     action_params = Map.drop(params, @wrapper_keys ++ @map_keys ++ [:__cascade__, "__cascade__"])
+    execution_params = ActionTool.convert_params_using_schema(action_params, mod.schema())
```
jfayad, line 265: **[medium] Make existing production map schemas compatible with validated execution**

This converter normalizes top-level NimbleOptions keys but leaves nested maps unchanged. Existing `ToUtcDateTime.delay` (`to_utc_datetime.ex:34–38`) and `ScheduleAction.params` (`schedule_action.ex:20`) still use `type: :map`, which NimbleOptions interprets as atom-key-only. Persisted workflow JSON therefore no longer reaches these actions.

Reproduced through `create_and_start_run/2`: `ToUtcDateTime` with `{"delay":{"amount":15,"unit":"minutes"}}` fails with `expected atom, got: "amount"`. The normalized ScheduleAction input similarly rejects `{"params":{"data":"hello"}}`. Both implementations already accept string keys; switching from direct execution introduces the rejection.

Repair the owning production Action schemas/normalization while retaining validated execution and safe arbitrary domain keys. Add real JSONB/workflow contract tests for both actions; their current direct `run/2` tests do not cover the newly enforced boundary.

jfayad:
### [medium] Record the required Action-reuse assessment and rollout ownership

The direction is appropriate: reuse Jido for validated execution rather than maintain parallel retry/timeout logic. However, I could not find the implementation's Action-reuse assessment or linked Beadwork/rollout issue in this PR, #706, or the relevant local tracking search. The review task `zaq-znt` is not implementation evidence.

`docs/action-reuse.md` requires concrete candidates, ownership/contract, execution and permission boundaries, and verification before approval. Please record those decisions for validated workflow execution and the new approval-control lifecycle, including why the policy/outcome/control helpers remain workflow-local.

The new `docs/services/workflows.md` rollout section also explicitly requires inventorying authorable definitions and nonterminal snapshots, reviewing nested map/Condition compatibility, coordinating old/new workers, and providing a compatible rollback. Link an owned rollout issue and record the compatibility disposition/validation plan; attach target-environment inventory before deployment. This is a request for evidence and accountable rollout planning, not a claim that a deployment has already violated the requirement.

Please also record required reloaded approval/rejection verification and the rationale for whether additional E2E work is needed. Existing CI browser checks are green, but are not evidence of deployed snapshot compatibility.

jfayad (changes requested):
## Verdict: request changes

Reviewed all three commits in the 33-file diff at `0833c57b`, against base `6e2650e6`, with #706 and the description/comments of related #761.

### Issue closure and quality

The narrow **#706 schema/default-handling defect is addressed**: the new EncodeBase64 regression exercises the real workflow boundary and asserts the persisted result. Reusing Jido execution, retaining submitted audit input, typed approval metadata, and property tests for suspension/control separation are meaningful improvements. **#761 is explicitly separate and is not resolved here.**

However, the PR is not merge-ready. Five reproduced code findings remain: silent loss of a pending-approval map item, production-action input regressions, incorrect published tool schema, and two driver lifecycle/recovery failures. Correct these before using this PR to close #706. The separate general comment requests required reuse/rollout evidence.

### Findings

**1 high, 5 medium, 0 low; 5 inline comments and 1 general conversation comment.** Five code findings were verified with six supplemental ExUnit cases outside the worktree; all six fail their expected contracts on this head. Two cases cover the same production-map-schema finding. No implementation or committed-test edits were made.

### Validation

- Focused nine-file suite: **224 tests, 4 properties, 0 failures**.
- `mix test test/zaq/engine/workflows test/zaq/agent/tools/workflow`: **977 tests, 19 properties, 0 failures**.
- `mix precommit`: formatting, strict Credo (1,298 files; no issues), and warnings-as-errors compilation passed. Test phase: **48 doctests, 134 properties, 8,475 tests, 5 failures, 1 skipped (86 excluded)**. Failures: one onboarding `render_async` timeout and four distributed-peer failures in `ConfidentialEventPeerTest`/`AuthRateLimiterPeerTest`. These are outside the diff; base-branch attribution was not established. The runner also timed out during teardown after printing the summary. **The final gate is not green.**
- Existing CI reports passing native Postgres/ParadeDB tests, format/Credo/docs and Chromium/Firefox/WebKit E2E; gitleaks is skipped. Browser E2E and numerical coverage were not rerun locally. `mix q` was not run because it mutates formatting/lockfiles; this review used the non-mutating precommit checks above.

### Coverage

Checked Jido/Runic/NimbleOptions seams, JSONB normalization, defaults/hooks, output contracts, retry/timeout ownership, map/fork accounting, approval/replay, permission context, lifecycle events, assertions/fixtures, documentation, reuse/duplication and project conventions. No BO/UI implementation changes are in scope. No independent new performance or duplication defect was established. Worktree remains clean.

lib/zaq/engine/workflows/workflow_run_agent.ex
jfayad: Re-evaluated at `84aea60e`: **addressed**. `RunWatcher.done/1` is no longer in an unconditional `after`; unhandled internal exits leave the watcher armed. The new internal-Runic-exit regression in `orphaned_run_recovery_test.exs` passed in the current-head suite (1,035 tests / 21 properties before local review-test additions). The separate existing RunWatcher suite also passed in isolation (8 tests). Resolving this thread.

lib/zaq/engine/workflows/step_runner.ex
jfayad: Re-evaluated at `84aea60e`: **addressed**. The owning ToUtcDateTime and ScheduleAction schemas now accept nested JSONB domain inputs through Zoi; their new persisted-run contract tests exercise the real StepRunner/Jido boundary. They also check delay constraints, generated tool schemas, and preservation of arbitrary nested scheduled params. The expanded NotifyPerson contract tests passed as well. Resolving this thread.

lib/zaq/agent/tools/workflow/condition.ex
jfayad: Re-evaluated at `84aea60e`: **addressed**. Condition now declares structured Zoi condition objects. The added assertions cover generated input/output schemas, published defaults, real `Tool.execute_action/3` arguments, and JSONB-reloaded workflow execution, rather than only direct validation. These tests passed in the current-head workflow/tool suite. Resolving this thread.

lib/zaq/engine/workflows/workflow_run_agent.ex
jfayad: Re-evaluated at `84aea60e`: **partially addressed; keeping open**. `LifecycleNotifier` contains `run.completed` / `run.waiting` failures, but `Workflows.update_run/3` still performs `Repo.update() |> broadcast_run()`, and `dispatch_async/2` calls NodeRouter without containment (`workflows.ex:704–708, 1750–1755, 1779–1785`). A raise on the UI event `{:broadcast, topic, {:run_updated, %{status: "completed"}}}` happens after commit and reaches the same stale-struct rescue. Reproduced for both completed and waiting: persisted status becomes **`interrupted`** in each case.

I added two local, uncommitted regression cases to `workflow_run_agent_test.exs` (`"#{status} state survives a post-commit UI broadcast failure"`); both currently fail at the persisted-status assertion. The tests assert that the database write happened before injecting the failure. Please contain the UI broadcast boundary too and/or make interruption a race-safe transition from the current persisted active state. This is the remaining path of the original finding, not a duplicate new finding.

lib/zaq/engine/workflows/map_node_builder.ex
jfayad: Re-evaluated at `84aea60e`: **partially addressed; keeping open**. The paused-approval recovery is fixed and its regression passes. However, raising in `handle_fork_result/3` does not durably propagate an unrecorded outer error: this code runs inside a native Runic Step, whose executor catches the exception and returns a failed Runnable. The driver applies it without writing a failed StepRun, then finalizes solely from persisted rows.

Confirmed by extending the same inconsistent approval/completed-cursor fixture already used in `map_node_builder_test.exs` with an independent successful leaf: `Workflows.start_run/1` returns **`status: "completed"`** while the approval remains pending. The new single-map test avoids false success only because no other terminal leaf completes; it does not establish failure propagation.

I added local, uncommitted test `"a completed sibling cannot hide an outer map error without a failed cursor"`; it currently fails (`completed` instead of failed/interrupted). Persist a fatal map/driver outcome or explicitly propagate failed map runnables to finalization; do not rely on the native-Step raise escaping Runic. This is a remaining case of the original failure-accounting finding, so no duplicate thread is being opened.

jfayad (changes requested):
## Re-review at `84aea60e` — request changes

Reviewed the four follow-up commits (21 changed files), surrounding contracts, prior findings, and #706/#761 context.

### Previous findings

**Confirmed addressed:**
- [Internal-exit watcher cleanup](https://github.com/www-zaq-ai/zaq/pull/765#discussion_r4083181582).
- [Condition generated tool schema](https://github.com/www-zaq-ai/zaq/pull/765#discussion_r4083181623).
- [Production nested JSONB Action schemas](https://github.com/www-zaq-ai/zaq/pull/765#discussion_r4083181619).

**Partially addressed; keep open:**
- [Finalized status overwrite](https://github.com/www-zaq-ai/zaq/pull/765#discussion_r4083181632): lifecycle-event dispatch is contained, but the UI broadcast after `Repo.update` still lets completed/waiting become interrupted.
- [Outer map-error accounting](https://github.com/www-zaq-ai/zaq/pull/765#discussion_r4083181652): paused approval recovery works, but Runic catches the new native-Step exception. An independent successful leaf still allows a completed run with a pending approval.

The [Action-reuse/rollout-evidence request](https://github.com/www-zaq-ai/zaq/pull/765#issuecomment-5762053950) remains outstanding.

**Resolution permission blocker:** GitHub MCP and CLI both rejected `resolveReviewThread` with “Resource not accessible by personal access token.” The three fixes are explicitly marked addressed in replies, but their threads remain open and need manual resolution or sufficient token permissions.

### Regression tests

As requested, added three useful reproductions to the existing project suite:
- `test/zaq/engine/workflows/workflow_run_agent_test.exs`: completed/waiting UI-broadcast failures.
- `test/zaq/engine/workflows/map_node_builder_test.exs`: completed sibling masking an outer map error.

These are **local, uncommitted/unpushed test-only changes**. All three retain failing expected-behavior assertions until the underlying defects are fixed. No production edits.

### Validation

- Pushed head before additions: **1,035 tests, 21 properties, zero failures** across workflow/tool/NotifyPerson suites.
- Changed test modules: **57 tests, exactly three new failures**.
- `mix q`: **passed** formatting, hooks, strict Credo, docs and compilation.
- `mix precommit`: checks passed; tests reported **7,232 tests, 159 properties, 42 doctests, exactly three failures, one skipped (86 excluded)**. Runner timed out during teardown after that summary. **Final gate remains red.**
- One existing RunWatcher sandbox queue timeout occurred in a separate broad run; isolated rerun passed 8/8, and final precommit did not reproduce it.

### Conclusion

Three medium findings are fixed. Outstanding: **one high and two medium**—two code threads plus one general comment. **No new distinct finding:** the additional failing cases belong to existing threads, so no duplicates were opened. Rechecked schema compatibility, permissions, map/replay behavior, lifecycle containment, tests, documentation, reuse and performance/duplication risks. #706 remains specifically fixed; #761 remains separate. Remaining correctness gaps prevent approval.

jfayad:
Pushed the three review regression tests in commit 3d4abf1e7 directly to this PR branch, as requested. Only the two existing test files changed; no production code changed.

Reproduce with:
```sh
mix test test/zaq/engine/workflows/map_node_builder_test.exs test/zaq/engine/workflows/workflow_run_agent_test.exs --seed 0
```

The three expected-behavior assertions currently fail: completed/waiting status is overwritten after a UI broadcast exception, and a successful sibling masks an outer map failure. They are intentionally not skipped or weakened so the fixes can make them pass. Commit and push hooks passed without bypasses.

PR #773, opened 2026-09-18
fix(ingestion): derive staged extensions from downloaded artifacts

### LINKED ISSUE #568: Ingestion: Docx conversion to markdown produces gibrish
When ingesting a docx file (in my example it was coming from my google drive connector) the resulting md is not the text content of the file but instead a badly decoded version of the docx file.

jat10:
fix #568


jfayad:
Two changes I’d recommend before merging:

1. **Avoid duplicating MIME/extension concerns.** `ArtifactType` already delegates to `MIME.extensions/1`, but it repeats some normalization that the MIME library performs, while `RecordSource` also contains related extension-safety logic. Please keep `MIME` as the authoritative mapping source and consolidate the ZAQ-specific normalization/safety policy in one place. If `RecordSource` remains the only consumer, private helpers there may be sufficient; otherwise a small shared wrapper is reasonable, but it should not duplicate MIME’s own behavior.

2. **Add a real conversion regression test for #568.** The current tests prove extension selection, byte preservation, and converter dispatch, but they do not convert an actual DOCX: several tests use arbitrary/fake DOCX bytes and fake converters that write predetermined Markdown. Please add a genuine DOCX fixture, mock only the provider download/materialization boundary, run the real local Python conversion, and assert that the resulting Markdown contains the fixture’s expected text and not ZIP/XML/binary gibberish. This requires no embedding or other external API.

Ideally, add the same fixture-based conversion contract for every offline format declared as supported (`.pdf`, `.docx`, `.pptx`, `.xlsx`, plus direct `.md`/`.csv` handling). Image formats can remain separate because their description path is API-backed. This would demonstrate that the PR fixes the reported failure rather than only proving that a `.docx` extension is selected.

jfayad (changes requested):
Two issues with this PR:
1. Spread concerns with duplication
2. Test assertions

jfayad:
@jat10 your tests are failing.

Moreover the update adds only the DOCX fixture-based conversion test, not equivalent PDF/PPTX/XLSX or direct MD/CSV contracts.

If these are supported, adding them in the tests scenarios is necessary, otherwise an issue with the remaining extensions should be created

jat10:
Agreed on both points.

The failing test is caused by the new real DOCX regression depending on crawler-ingest Python assets and dependencies that the normal CI environment does not provision. I opened #781 to make that Python environment deterministic in development and CI rather than weakening the regression test.

I also validated the remaining supported offline formats through the real `DocumentProcessor.prepare_file_chunks/2` path. PDF, DOCX, PPTX, XLSX, Markdown, and CSV all pass without requiring production conversion changes, so I’ll add equivalent fixture-based contract coverage to this PR.

PDF/DOCX/PPTX/XLSX will rely on the real-Python test boundary from #781, while Markdown and CSV can stay in the normal ExUnit suite.

During that validation, I also found an unrelated same-basename collision where converted documents can overwrite an existing sibling `.md` file. I opened #782 for that instead of expanding #568 further.

PR #776, opened 2026-09-19
feat(credentials): invalidate agent runtimes on auth changes

jfayad:


lib/zaq/agent/factory.ex
```diff
@@ -145,19 +150,68 @@ defmodule Zaq.Agent.Factory do
...
+        %System.AIProviderCredential{} = attached -> attached
+        _ -> load_ai_credential(configured_agent.credential_id)
+      end
+
+    case credential do
+      nil ->
+        {:ok, nil}
+
+      %{connect_credential_id: id} when is_integer(id) ->
+        connect_module = Keyword.get(opts, :connect_module, Connect)
```
jfayad: Connect belongs to the Engine, no direct function calling should occur from Agent to Engine

